PR #35, opened 2018-09-22
Fixed lint warnings and added golint,go vet to build script

vramakrishnan:


scripts/build.sh
```diff
@@ -2,8 +2,26 @@
...
     chmod -R 777 bin
-    echo -e "\e[42;97m SUCCESS \e[0m"
+    echo -e "\e[42;97m BUILD SUCCESS \e[0m"
 else
-    echo -e "\e[101;97m FAILED \e[0m"
+    echo -e "\e[101;97m BUILD FAILED \e[0m"
     exit 1
 fi
+
+echo -e "\e[44;97m Running go lint ... \e[0m"
```
asridharan: We should run the linter before the compilation? Linters are generally faster. The compilations would take longer. So we can potentially error out earlier.

asridharan:
Thinking abou this a bit more. We can probably take it in a separate PR:  
* We should have a separate target for linter.
* Also, instead of invoking this shell script we should be executing these commands directly in the docker run commands. The failure codes for golang should automatically propoagte to CMake.

Will create a separate TechDebt PR to improve the build system.

scripts/build.sh
vramakrishnan: Will do.

asridharan:
@vramakrishnan just fixed the conflicts. Will merge once the checks pass.

asridharan:
@vramakrishnan looks like this branch still has conflicts. Could I request you to rebase the branch again?

PR #43, opened 2018-09-26
Controller tests

asridharan:
Added part I of the unit-tests for `k8scontext` package.

    The Kubernetes go-client provides a package for mocking the Kubernetes API server. We use this mock framework to introduce Ginkgo/Gomega based test framework for the package. This is the first part of a series of commits to introduce unit-tests for this package. In this commit we only test the ability of the controller to listen to CRUD operations to Ingress Resources. We also test that it only reacts to CRUD events on Ingress Resources that have the right annotations set.


asridharan:
@vramakrishnan  could you take a look?

pkg/k8scontext/k8scontext_test.go
```diff
@@ -0,0 +1,78 @@
...
+						ServiceName: "hello-world",
+						ServicePort: intstr.IntOrString{
+							Type:   intstr.Int,
+							IntVal: 80,
+						},
+					},
+				},
+			}
+
+			// l.Println("Running the ingress test !!")
```
vramakrishnan: Please remove unused code.

pkg/k8scontext/k8scontext_test.go
```diff
@@ -0,0 +1,78 @@
...
+		// Create the "test-ingress-controller" namespace.
+		// We will create all our resources under this namespace.
+		ns := &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: ingressNS,
+			},
+		}
+		_, err := k8sClient.CoreV1().Namespaces().Create(ns)
+
+		Expect(err).Should(BeNil(), "Unable to create the mock Kuberentes client: %v", err)
```
vramakrishnan: nit: typo kuberentes and should the message be more like "unable to create namespace"

pkg/k8scontext/k8scontext_test.go
```diff
@@ -20,6 +21,49 @@ var _ = Describe("K8scontext", func() {
...
+		},
+	}
+
+	// Create the Ingress resource.
+	ingress := &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ingressName,
+			Namespace: ingressNS,
+			Annotations: map[string]string{
+				k8scontext.IngressClass: k8scontext.IngressControllerName,
```
vramakrishnan: Shouldn't it be IngressClass instead ?

pkg/k8scontext/k8scontext_test.go
```diff
@@ -0,0 +1,190 @@
...
```
vramakrishnan: Shouldn't this be IngressClass instead ?

pkg/k8scontext/k8scontext_test.go
```diff
@@ -0,0 +1,190 @@
...
+	BeforeEach(func() {
+		// Create the mock K8s client.
+		k8sClient = testclient.NewSimpleClientset()
+
+		_, err := k8sClient.CoreV1().Namespaces().Create(ns)
+
+		Expect(err).Should(BeNil(), "Unable to create the mock Kuberentes client: %v", err)
+
+		_, err = k8sClient.Extensions().Ingresses(ingressNS).Create(ingress)
+		Expect(err).Should(BeNil(), "Unabled to create ingress resource due to: %v", err)
```
vramakrishnan: nit: typo kuberentes and err msg can be more specific here IMO - "unable to create namespace"

pkg/k8scontext/k8scontext_test.go
```diff
@@ -28,51 +72,119 @@ var _ = Describe("K8scontext", func() {
...
+
+			ingressListInterface := ctxt.Caches.Ingress.List()
+			Expect(len(ingressListInterface)).To(Equal(1), "Expected to have a single ingress in the cache but found: %d ingresses", len(ingressListInterface))
+
+			// Retrive the ingresses learnt by the controller.
+			testIngresses := ctxt.GetHTTPIngressList()
+			Expect(len(testIngresses)).To(Equal(1), "Expected to have a single ingress in the k8scontext but found: %d ingresses", len(testIngresses))
+
+			// Make sure the ingress we got is the ingress we stored.
+			Expect(testIngresses[0]).To(Equal(ingress), "Expected to retrieve the same ingress that we inserted, but it seems we found the following ingress: %v", testIngresses[0])
```
vramakrishnan: question: does v1beta1.Ingress satisfy the GomegaMatcher  interface ? (making it possible to with Equal())

pkg/k8scontext/k8scontext_test.go
```diff
@@ -28,51 +72,119 @@ var _ = Describe("K8scontext", func() {
...
+
+			// Retrive the ingresses learnt by the controller.
+			testIngresses := ctxt.GetHTTPIngressList()
+			Expect(len(testIngresses)).To(Equal(1), "Expected to have a single ingress in the k8scontext but found: %d ingresses", len(testIngresses))
+			// Make sure the ingress we got is the ingress we stored.
+			Expect(testIngresses[0]).To(Equal(ingress), "Expected to retrieve the same ingress that we inserted, but it seems we found the following ingress: %v", testIngresses[0])
+
+		})
+
+		It("Should be able to follow modifications to the Ingress Resource.", func() {
```
vramakrishnan: "It" descriptions needs to be fixed to indicate that it is a "delete" test.

pkg/k8scontext/k8scontext_test.go
```diff
@@ -28,51 +72,119 @@ var _ = Describe("K8scontext", func() {
...
+			Expect(err).Should(BeNil(), "Unable to retrieve stored ingresses resource due to: %v", err)
+			Expect(len(ingresses.Items)).To(Equal(0), "Expected to have no ingresses stored in mock K8s but found: %d ingresses", len(ingresses.Items))
+
+			// Due to the large sync time we don't expect the cache to be synced, till we force sync the cache.
+			// Start the informers. This will sync the cache with the latest ingress.
+			ctxt.Run()
+
+			ingressListInterface := ctxt.Caches.Ingress.List()
+			// There should still be only one ingress resource.
+			Expect(len(ingressListInterface)).To(Equal(0), "Expected to have no ingress in the cache but found: %d ingresses", len(ingressListInterface))
```
vramakrishnan: Do we have to wrap these any of these test assertions with Eventually (i.e retries) to handle timing issues ?

pkg/k8scontext/k8scontext_test.go
```diff
@@ -28,51 +72,119 @@ var _ = Describe("K8scontext", func() {
...
+			testIngresses := ctxt.GetHTTPIngressList()
+			Expect(len(ingressListInterface)).To(Equal(0), "Expected to have no ingress in the k8scontext but found: %d ingresses", len(testIngresses))
+		})
+
+		It("Should be following Ingress Resource with Application Gateway specific annotations only.", func() {
+			nonAppGWIngress := &v1beta1.Ingress{}
+			deepcopy.Copy(nonAppGWIngress, ingress)
+			nonAppGWIngress.Name = ingressName + "123"
+			// Change the `Annotation` so that the controller doesn't see this Ingress.
+			nonAppGWIngress.Annotations[k8scontext.IngressClass] = k8scontext.IngressControllerName + "123"
```
vramakrishnan: Same comment as previous one, Should this be IngressClass ?

pkg/k8scontext/k8scontext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			ctxt.Run()
+
+			ingressListInterface := ctxt.Caches.Ingress.List()
+			// There should still be only one ingress resource.
+			Expect(len(ingressListInterface)).To(Equal(0), "Expected to have no ingress in the cache but found: %d ingresses", len(ingressListInterface))
+
+			// Retrive the ingresses learnt by the controller.
+			testIngresses := ctxt.GetHTTPIngressList()
+			Expect(len(ingressListInterface)).To(Equal(0), "Expected to have no ingress in the k8scontext but found: %d ingresses", len(testIngresses))
+		})
```
vramakrishnan: AFAIK, "It"s can be executed in any order and if this runs earlier, then some of the Modify tests the precedes this can fail. Something to watch out for.

pkg/k8scontext/k8scontext_test.go
asridharan: Yeah. Its pretty intelligent. I haven't looked at the internals, but I think it uses the `reflect` package to do deep inspection to equate properties of the two objects. Similar to how `deepcopy` works using the reflect package.

pkg/k8scontext/k8scontext_test.go
asridharan: There shouldn't be any retries here since the calls are either against the Mock or against the cache. These are there just invoking different go methods rather than making actual network/sock calls.

pkg/k8scontext/k8scontext_test.go
asridharan: The `It` depends on the `BeforeEach` and not any of the other `Its`, the order of the execution of `It` should not matter.

njzhangyifei:
Other than comments from @vramakrishnan , I think it looks good.

asridharan:
@vramakrishnan can you take another look?

PR #48, opened 2018-10-01
clean NewAppGwIngressController

dvrkps:


pkg/controller/controller.go
```diff
@@ -34,12 +34,12 @@ type AppGwIngressController struct {
 // NewAppGwIngressController constructs a controller object.
 func NewAppGwIngressController(kubeclient kubernetes.Interface, appGwClient network.ApplicationGatewaysClient, appGwIdentifier appgw.Identifier, k8sContext *k8scontext.Context) *AppGwIngressController {
 	controller := &AppGwIngressController{
+		appGwClient: appGwClient,
 		appGwIdentifier:  appGwIdentifier,
 		k8sContext:       k8sContext,
 		k8sUpdateChannel: k8sContext.UpdateChannel,
 	}
 	controller.eventQueue = NewEventQueue(controller.processEvent)
```
vramakrishnan: We could also move eventQueue initialization up.

pkg/controller/controller.go
```diff
@@ -34,12 +34,12 @@ type AppGwIngressController struct {
 // NewAppGwIngressController constructs a controller object.
 func NewAppGwIngressController(kubeclient kubernetes.Interface, appGwClient network.ApplicationGatewaysClient, appGwIdentifier appgw.Identifier, k8sContext *k8scontext.Context) *AppGwIngressController {
```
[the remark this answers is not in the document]
dvrkps: @asridharan @vramakrishnan Done.

pkg/controller/controller.go
dvrkps: @vramakrishnan i think that is not true. https://play.golang.org/p/-VciWJJPDgz

vramakrishnan: You are right. Callback needs to be initialized in next step, the way it is today.

asridharan:
@dvrkps `go vet` is failing with the following error:
```
https://travis-ci.org/Azure/application-gateway-kubernetes-ingress/builds/435772383#L952
```
You need remote the `Kubernetes` package from the import.

dvrkps:
@asridharan Done.

PR #53, opened 2018-10-05
Added unit-test for the `appgw` package.

asridharan:
Have added tests for testing the generation of an application gateway configuration through the `ConfigBuilder` interface . In this commit we are testing only the HTTP ingress rules.

### DIFF
--- a/pkg/appgw/appgw_test.go
+++ b/pkg/appgw/appgw_test.go
@@ -1,0 +1,293 @@
+package appgw
+
+import (
+	go_flag "flag"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/api/core/v1"
+	"k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+	testclient "k8s.io/client-go/kubernetes/fake"
+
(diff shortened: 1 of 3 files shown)


pkg/appgw/appgw_test.go
```diff
@@ -0,0 +1,286 @@
...
+						IP: endpoint2,
+					},
+					{
+						IP: endpoint3,
+					},
+				},
+				Ports: []v1.EndpointPort{
+					{
+						Name:     "database",
+						Port:     backendPort,
```
vramakrishnan: Nit:  "frontend" vs "database" naming, should be fixed IMO.

pkg/appgw/appgw_test.go
```diff
@@ -0,0 +1,286 @@
...
+		ctxt = k8scontext.NewContext(k8sClient, ingressNS, 1000*time.Second)
+		Expect(ctxt).ShouldNot(BeNil(), "Unable to create `k8scontext`")
+
+		// Initialize the `ConfigBuilder`
+		configBuilder = NewConfigBuilder(ctxt, &Identifier{}, &network.ApplicationGatewayPropertiesFormat{})
+
+		builder, ok := configBuilder.(*appGwConfigBuilder)
+		Expect(ok).Should(BeTrue(), "Unable to get the more specific configBuilder implementation")
+
+		builder.appGwConfig.FrontendIPConfigurations = &[]network.ApplicationGatewayFrontendIPConfiguration{
```
asridharan: Explain why we need this.

asridharan:
@vramakrishnan can you please take another look?

Issue #32, opened 2018-09-20
Use sync.map insteadof `ThreadsafeMultiMap`


asridharan:
**Is your feature request related to a problem? Please describe.**
Since we are using Golang 10.3 we should be using the built in sync.map instead of using the `ThreadsafeMultiMap` package that we introduced. This would help simplify our code.

asridharan:
Closing this out due to inactivity.

PR #55, opened 2018-10-06
Use sync.Map instead of custom ThreadsafeMultiMap implementation

Xjs:
Could resolve https://github.com/Azure/application-gateway-kubernetes-ingress/issues/32 – not sure if and why it is advantageous at this point to use sync.Map. Can't use a sync.Map containing an UnorderedSet, because set access might still not be synchronized.

(I'm happy to open a discussion at this point – but I wanted to get at least one PR out while on the #gofrm meetup. ;))

### DIFF
--- a/pkg/utils/threadsafemultimap.go
+++ b/pkg/utils/threadsafemultimap.go
@@ -11,100 +11,73 @@ import "sync"
 type ThreadsafeMultiMap interface {
 	Insert(key interface{}, value interface{})
 	Clear(key interface{})
-	Erase(key interface{}) bool
-	EraseValue(value interface{}) bool
+	Erase(key interface{})
+	EraseValue(value interface{})
 	ContainsPair(key interface{}, value interface{}) bool
 	ContainsValue(value interface{}) bool
 }
 
 type threadsafeMultiMap struct {
-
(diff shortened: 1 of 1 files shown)


njzhangyifei:
> I had a related question, requirement for multi-map (key with multiple values) to start with.

`ThreadsafeMultimap` contains mappings between a key and an unordered set of values. I was using it to maintain `ingKey -> secKey[]` mappings. This way, we will be able to achieve the minimum number of certificate conversions while saving their relationship.

`ingressSecretsMap` in [this file](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/pkg/k8scontext/context.go)

asridharan:
@Xjs just to elaborate on @njzhangyifei comment above and as you mentioned in your opening remark we need to store the `unordered` set in the `ThreadSafeMultiMap` implementation to emulate the Multi-Map, which makes it  a bit tricky to use the `sync.Map`. @vramakrishnan and me had an internal discussion and the right approach here might be to use a `sync.Map` both fore the `Multi-Map` and to represet the `unordered` set.

asridharan:
@Xjs based on the discussion above closing this PR for the time being since looks like its not being worked on.

PR #56, opened 2018-10-08
Add goimports to the build script

vramakrishnan:


asridharan:
@vramakrishnan need to rebase.

Issue #31, opened 2018-09-20
Support AKS RBAC enabled clusters


asridharan:
**Is your feature request related to a problem? Please describe.**
We should be able to run the ingress controller on RBAC enabled AKS clusters.

**Describe the solution you'd like**
We need to add templates to the Helm chart to setup a service principal and give the service principal the right permissions to get access to the resources.
https://github.com/Azure/aad-pod-identity/blob/master/deploy/infra/deployment-rbac.yaml

vramakrishnan:
Support added in https://github.com/Azure/application-gateway-kubernetes-ingress/pull/57

PR #57, opened 2018-10-17
Support Rbac enabled cluster

akshaysngupta:
https://github.com/Azure/application-gateway-kubernetes-ingress/issues/31


deploy/azuredeploy.json
```diff
@@ -307,6 +307,13 @@
             "metadata": {
                 "description": "A CIDR notation IP for Docker bridge."
             }
+        },
+        "aksEnableRBAC": {
+            "type": "bool",
+            "defaultValue": false,
```
vramakrishnan: Is there a reason why we want RBAC to be disabled by default ? 
I would suggest to make it default. 

Please see this cluster hardening guide on using RBAC :
https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster

asridharan: That is a good question. I think this point did come up in discussion with the AKS folks and if I understand correctly the default for AKS right now is RBAC disabled by default and would therefore make sense to keep the default value in the helm chart to be disabled as well.

PR #80, opened 2018-12-18
Update azuredeploy.json with a default Kubernetes version available t…

knee-berts:
…o AKS


asridharan:
Thanks for the PR @knee-berts

PR #84, opened 2018-12-31
[Dotnotmerge] Annotation backend-path-prefix

akshaysngupta:


pkg/annotations/ingress_annotations.go
```diff
@@ -0,0 +1,51 @@
...
+	extensions "k8s.io/api/extensions/v1beta1"
+)
+
+const (
+	// Prefix AppGw Prefix
+	Prefix = "appgw.ingress.kubernetes.io"
+
+	// BackendPathPrefix defines the value for Path which should be used as a prefix for all HTTP requests.
+	// Null means no path will be prefixed. Default value is null.
+	BackendPathPrefix = "appgw.ingress.kubernetes.io/backend-path-prefix"
```
vramakrishnan: Is this Key or Value ? Can we update the comments.
It is used as Key in in Ingress.Annotations Map and also as Value for Path in ApplicationGatewayBackendHTTPSettingsPropertiesFormat.

vramakrishnan: Also please use the "Prefix" in subsequent definition such as in BackendPathPrefix .

pkg/appgw/appgw_test.go
```diff
@@ -55,7 +56,8 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
 			Name:      ingressName,
 			Namespace: ingressNS,
 			Annotations: map[string]string{
-				k8scontext.IngressClass: k8scontext.IngressControllerName,
+				annotations.IngressClassKey:   annotations.IngressControllerName,
+				annotations.BackendPathPrefix: "/",
```
vramakrishnan: Please add test to verify/read the Annotations settings in - 
"Context("Tests Application Gateway Configuration", func()"

PR #92, opened 2019-02-10
Backend path prefix

asridharan:
Introduced the backend prefix as an annotation.

Have run E2E tests with multiple ingress resources, to test the backend path prefix annotation.

### DIFF
--- a/pkg/annotations/ingress_annotations.go
+++ b/pkg/annotations/ingress_annotations.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	// BackendPathPrefixKey defines the value for Path which should be used as a prefix for all HTTP requests.
+	// BackendPathPrefixKey defines the key for Path which should be used as a prefix for all HTTP requests.
 	// Null means no path will be prefixed. Default value is null.
 	BackendPathPrefixKey = "appgw.ingress.kubernetes.io/backend-path-prefix"
 
--- a/pkg/annotations/ingress_annotations.go
+++ b/pkg/annotations/ingress_annotations.go
@@ -1,51 +1,37 @@
 package annotations
 
 import (
-	extensions "k8s.io/api/extensions/v1beta1"
+	"k8s.io/api/extensions/v1beta1"
 )
 
 const (
-	// Prefix AppGw Prefix
-	Prefix = "appgw.ingress.kubernetes.io"
-
-	// BackendPathPrefix defines the value for Path which should be used as a prefix for all HTTP requests.
+	// BackendPathPrefixKey defines the value for Path which should be used as a prefix for all HTTP requests.
 	// Null means no path will be prefixed. Default value is null.
-	BackendPathPrefix = "appgw.ingress.kubernetes.io/backend-path-prefix"
+	BackendPathPrefixKey = "appgw.ingress.kubernetes.io/backend-path-prefix"
 
-	// IngressClass defines the key of the annotation which needs to be set in order to specify
+	// IngressClassKey defines the key of the annotation which needs to be set in order to specify
 	// that this is an ingress resource meant for the application gateway ingress controller.
 	IngressClassKey = "kubernetes.io/ingress.class"
 
-	// IngressControllerName defines the value of the `IngressClass` annotation that will tell the ingress controller
+	// ApplicationGatewayIngressClass defines the value of the `IngressClassKey` annotation that will tell the ingress controller
 	// whether it should act on this ingress resource or not.
-	IngressControllerName = "azure/application-gateway"
+	ApplicationGatewayIngressClass = "azure/application-gateway"
 )
 
-// IngressAnnotations helper
-type IngressAnnotations struct {
-	annotations map[string]string
-}
-
-// FromIngress Annotations
-func FromIngress(ing *extensions.Ingress) *IngressAnnotations {
-	return &IngressAnnotations{ing.Annotations}
-}
-
-// OverrideBackendPath override path
-func (ing *IngressAnnotations) BackendPathPrefix() string {
-	val, ok := ing.annotations[BackendPathPrefix]
-	if !ok {
-		return ""
-	}
-
+// BackendPathPrefix override path
+func BackendPathPrefix(ing *v1beta1.Ingress) string {
+	val, _ := ing.Annotations[BackendPathPrefixKey]
 	return val
 }
 
 // IngressClass ingress class
-func (ing *IngressAnnotations) IngressClass() string {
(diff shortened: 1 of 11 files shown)


pkg/annotations/ingress_annotations.go
```diff
@@ -0,0 +1,37 @@
+package annotations
+
+import (
+	"k8s.io/api/extensions/v1beta1"
+)
+
+const (
+	// BackendPathPrefixKey defines the value for Path which should be used as a prefix for all HTTP requests.
```
vramakrishnan: Shouldn't the comment be " BackendPathPrefixKey defines the key of the annotation which needs to be set in order to specify the prefix for all HTTP requests" ?  (Key vs Value clarification)

For all appgw annotations that we will add in future, maybe good to create a annotation base definition later on - "appgw.ingress.kubernetes.io"

pkg/appgw/internaltypes.go
```diff
@@ -63,8 +65,8 @@ func getResourceKey(namespace, name string) string {
 	return fmt.Sprintf("%v/%v", namespace, name)
 }
 
-func generateHTTPSettingsName(serviceName string, servicePort string, backendPortNo int32) string {
-	return fmt.Sprintf("k8s-%v-%v-bp-%v", serviceName, servicePort, backendPortNo)
+func generateHTTPSettingsName(serviceName string, servicePort string, backendPortNo int32, ingress string) string {
+	return fmt.Sprintf("k8s-%v-%v-bp-%v-%s", serviceName, servicePort, backendPortNo, ingress)
```
vramakrishnan: We were trying to find all appgw's configured as ingress using "contains "k8s-" query in config and it gave more matches since "k8s-" was also used by other users in their configs. Maybe having something unique "k8s-appgw-" or the likes will help query these auto-generated configs when ingress controller is used.

akshaysngupta:
It would be good to add a readme listing all the annotations and their usage.

asridharan:
@vijayrkn addressed your comments. Need to add an annotations doc and add  a tutorial for this PR. @akshaysngupta  will address it in a separate PR.

pkg/annotations/ingress_annotations.go
asridharan: Will create a base annotation when add more.

asridharan:
@vramakrishnan @akshaysngupta PTAL. Have added documentation.

wesselkranenborg:
Please be carefull with this PR and test this good with longer host names as we are facing a bug in application gateway right now. The bug is that the AG becomes unhealthy when we have a hostname longer than X (still determining) characters. During this PR I see that the prefix changes from 'k8s-' to 'k8s-ag-ingress' and this will cause hitting this issue earlier. Just mentioning it here so that you are aware.

asridharan:
> Please be carefull with this PR and test this good with longer host names as we are facing a bug in application gateway right now. The bug is that the AG becomes unhealthy when we have a hostname longer than X (still determining) characters. During this PR I see that the prefix changes from 'k8s-' to 'k8s-ag-ingress' and this will cause hitting this issue earlier. Just mentioning it here so that you are aware.

@wesselkranenborg we don't generate hostname strings in the ingress-controller? These are just resource names and ARM doesn't have limitations right now on the length of resource names. So we should be fine here.

wesselkranenborg:
> we don't generate hostname strings in the ingress-controller? 
Indeed, that's also not the problem but the resource names are generated based on the hostname strings.

> These are just resource names and ARM doesn't have limitations right now on the length of resource names. So we should be fine here.
ARM indeed doesn't have limitations and the configuration also gets deployed to the application gateway. But then the application gateway is going to an unheathy state so no (zero) requests are being handled by the application gateway although the configuration looks good.

We are able to reproduce these issues in our environments right now. We mitigating this by choosing shorter hostnames and than everything works.

asridharan:
> > we don't generate hostname strings in the ingress-controller?
> > Indeed, that's also not the problem but the resource names are generated based on the hostname strings.
> 
> > These are just resource names and ARM doesn't have limitations right now on the length of resource names. So we should be fine here.
> > ARM indeed doesn't have limitations and the configuration also gets deployed to the application gateway. But then the application gateway is going to an unheathy state so no (zero) requests are being handled by the application gateway although the configuration looks good.
> 
> We are able to reproduce these issues in our environments right now. We mitigating this by choosing shorter hostnames and than everything works.

Sorry, didn't follow your response. Do you mean to say that the ingess controller is setting up the hostname for you?

asridharan:
> > we don't generate hostname strings in the ingress-controller?
> 
> Indeed, that's also not the problem but the resource names are generated based on the hostname strings.
> 
> > These are just resource names and ARM doesn't have limitations right now on the length of resource names. So we should be fine here.
> 
> ARM indeed doesn't have limitations and the configuration also gets deployed to the application gateway. But then the application gateway is going to an unheathy state so no (zero) requests are being handled by the application gateway although the configuration looks good.
> 
> We are able to reproduce these issues in our environments right now. We are mitigating this by choosing shorter hostnames and than everything works.
> 
> For this issues are we currently working together with microsoft support on a solution.

So seems like the issue is with setting up the hostname for a given listener rather than the resource names itself. This seems to be an orthogonal problem with AG that we will have to fix independent of the ingress controller.  So we should be good here.

wesselkranenborg:
> This seems to be an orthogonal problem with AG that we will have to fix independent of the ingress controller.

I agree with you on this but am just mentioning this because a longer prefix will cause faster hitting this bug. Which in the end will slow down the adoption of this ingress controller (despite the fact that it has nothing to do with the ingress controller). It's just FYI but you are now aware of this bug.

We are not going to upgrade to this new version until the bug in the AG is fixed. Otherwise, our cluster will not be reachable.

asridharan:
> > This seems to be an orthogonal problem with AG that we will have to fix independent of the ingress controller.
> 
> I agree with you on this but am just mentioning this because a longer prefix will cause faster hitting this bug. Which in the end will slow down the adoption of this ingress controller (despite the fact that it has nothing to do with the ingress controller). It's just FYI but you are now aware of this bug.
> 
> We are not going to upgrade to this new version until the bug in the AG is fixed. Otherwise, our cluster will not be reachable.

@wesselkranenborg I think I am missing your point here? How is this prefix related to the hostname that you are setting? I understand there seems to be a bug in AG with the hostname length, but you seem to be somehow implying that the prefix we are introducing is somehow related to the hostnames that you are setting in AG and this would speed up the time in which you would hit this bug? 

Want to understand the implication to see if we can fix any issues before they happen. Could you share any support case that you might already be working with Azure on this?

PR #106, opened 2019-02-21
Mark preview features as preview

RicardoNiepel:
Mark preview features as preview: application-gateway-kubernetes-ingress


asridharan:
@RicardoNiepel could you give some context as to why we need this PR? Is this a MSFT Azure requirement?

Issue #70, opened 2018-11-12
HttpListener is created for http when http is disabled


wesselkranenborg:
Currently there is always a httpListener created for both http and https even if http is disabled in the ingress

The solution i'd like is to skip the generation of the httpListener creation if http is disabled.

See also this TODO in the code https://github.com/Azure/application-gateway-kubernetes-ingress/blob/6a9ce466bd37133d569417905eca6ae912df55b5/pkg/appgw/httpListeners.go#L96

asridharan:
@wesselkranenborg make sense. Just to clarify, you want the HTTP listener to be disabled if we specify only HTTPS in the ingress right?

wesselkranenborg:
That is indeed what we want.

dan-jackson-github:
@asridharan can you indicate when the annotation support for disabling HTTP access will be available? Creating an HTTP endpoint by default really isn't ideal and means more manual work to do remember after deployment

robertprather:
I support all previous comments on this thread and would like to see this implemented ASAP. 

Using the application gateway to enable the HTTPS listener and perform SSL offload to the K8S cluster is working great, but leaving the HTTP listener enabled is causing us grief with our security team. We want the ingress controller to update the gateway and set the desired state without manual updates. 

An annotation would work, suggest 2 possible ones:

1. ssl-require: if this annotation is present, would disable the HTTP listener if HTTPS is specified in the ingress
2. ssl-redirect: if this annotation is present, would keep both the HTTP and HTTPS listeners present on the gateway, but apply a redirect rule for HTTP -> HTTPS (k8s--80-rr rule would redirect k8s--80-fl listener to k8s--443-fl listener)

asridharan:
@wesselkranenborg @robertprather have a partial fix for this problem in
https://github.com/Azure/application-gateway-kubernetes-ingress/pull/120

Its partial since even after this PR we won't have support for SSL-redirect on HTTP listeners. Will work on that after this PR. Wanted to check if it would help resolve this issue?

asridharan:
@wesselkranenborg @dan-jackson-github @robertprather can you try helm chart 0.3.0 which has the fix for this issue?

Thanks,
Avinash

asridharan:
Fixed by #120

robertprather:
I've tried helm chart 0.3.0 and it works as expected - no http listener created. Rob

PR #120, opened 2019-03-26
Modified creation of HTTP listener.

asridharan:
We were creating HTTP listeners for ingress spec that had TLS specified. Modifed the listener creation to create HTTP listeners only if the TLS spec is not specified.


asridharan:
@vramakrishnan  @akshaysngupta PTAL.

This fixes #70 

Currently when we specify a TLS spec in ingress we end up creating an HTTP listener for the TLS spec as well. The idea was to introduce an annotation for SSL-redirect at a later stage which hasn't happened yet (should be straight forward to add). But in the interim introducing this incremental change (which will be later on used for ssl-redirect annotation as well) to not create an HTTP listener for an ingress that has specified a TLS spec.

PR #122, opened 2019-03-29
TLS tests

asridharan:
* Introduced infrastructure to run TLS tests.
* Introduced a unit-test to test ingress spec with TLS.
* Fixed flaky tests. We were relying on the ordering of the HTTP listeners and backend address pools in the application gateway config in the tests, which were resulting in flaky tests. Have used the `ConsistOf` and `ContainsElement` calls to remove the requirement of ordering of these entries in the tests.


asridharan:
@akshaysngupta @vramakrishnan can you PTAL?

asridharan:
@vramakrishnan could you sign off again? 

Thanks,
Avinash

vramakrishnan:
LGTM

PR #227, opened 2019-05-31
Removing forEach and ToSlice() from unordered set

rheaj99:
Removed forEach method from unordered set


akshaysngupta:
@rheaj99 Can you also add tests for `Union` and `Intersection` functions ?

PR #361, opened 2019-07-08
Add Istio Endpoints

rheaj99:
This PR defines a function that goes through all the http route rules and destinations for a given Istio virtual service, and uses the existing GetEndpointsByService function to get the endpoints for each destination service.


pkg/istio/controller/endpoints.go
```diff
@@ -1,11 +1 @@
 package controller
```
akshaysngupta: please remove this file.

PR #367, opened 2019-07-10
Get annotated gateways

rheaj99:
This PR adds an annotation to Istio Gateway to indicate that they should be handled by AGIC and creates a function to get all annotated gateways

### DIFF
--- a/pkg/annotations/ingress_annotations.go
+++ b/pkg/annotations/ingress_annotations.go
@@ -8,6 +8,7 @@ package annotations
 import (
 	"strconv"
 
+	"github.com/knative/pkg/apis/istio/v1alpha3"
 	"k8s.io/api/extensions/v1beta1"
 
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/annotations/ingress_annotations.go
```diff
@@ -40,13 +41,17 @@ const (
 	// that this is an ingress resource meant for the application gateway ingress controller.
 	IngressClassKey = "kubernetes.io/ingress.class"
 
+	// IstioGatewayKey defines the key of the annotation which needs to be set in order to specify
+	// that this is sa gateway meant for the application gateway ingress controller.
```
akshaysngupta: typo

pkg/annotations/ingress_annotations.go
```diff
@@ -40,13 +41,17 @@ const (
...
+
 	// ApplicationGatewayIngressClass defines the value of the `IngressClassKey` annotation that will tell the ingress controller
 	// whether it should act on this ingress resource or not.
 	ApplicationGatewayIngressClass = "azure/application-gateway"
 
 	// ApplicationGatewayIstioIngressClass defines the value of the `IngressClassKey` annotation that will tell the ingress controller whether
-	// this ingress resource is from Istio or not.
-	ApplicationGatewayIstioIngressClass = "istio-ingress"
+	// this gateway should be handled by AGIC or not
+	ApplicationGatewayIstioIngressClass = "controls-app-gw"
```
akshaysngupta: can we use the same ingress class for Istio ?

pkg/k8scontext/context.go
```diff
@@ -347,6 +347,18 @@ func (c *Context) GetEndpointsForVirtualService(virtualService v1alpha3.VirtualS
 	}
 }
 
+// GetGateways returns all Istio Gateways that are annotated.
+func (c *Context) GetGateways(annotation string) []*v1alpha3.Gateway {
```
akshaysngupta: `annotation string` are we using this anywhere ?

pkg/k8scontext/context.go
```diff
@@ -347,6 +347,18 @@ func (c *Context) GetEndpointsForVirtualService(virtualService v1alpha3.VirtualS
...
+	annotatedGateways := make([]*v1alpha3.Gateway, 0)
+	for _, gateway := range c.ListIstioGateways() {
+		annotated, _ := annotations.IsIstioGatewayIngress(gateway)
```
akshaysngupta:
```go
if annotated, _ := annotations.IsIstioGatewayIngress(gateway); annotated {
    annotatedGateways = append(annotatedGateways, gateway)
}
```

pkg/annotations/ingress_annotations.go
rheaj99: I'm not sure - the Gateway I'm playing around with right now doesn't have an ingress class annotation, so I'm not sure if it's necessary, but I can add that in later if it is.

PR #408, opened 2019-07-24
s/Application Gateway/App Gateway/g

delqn:
This is a no-op PR renaming all string literal and comment occurrences of `Application Gateway` to `App Gateway`

This is mainly for consistency - we tend to use these two interchangeably. In logs (which we have many of) would be great to consistently use the shorter form.


akshaysngupta:
Should we make everything `Application Gateway` instead in full ?

delqn:
shelving for later

PR #409, opened 2019-07-24
errors: Make errors.New() singletons - into errors.go

delqn:
The goal of this PR is to establish a pattern where we instantiate types of errors once, and then reuse these globals.

This affords us easier error checking / error comparison.

Example (https://play.golang.org/p/S0Y2_0qBsbi):
```golang
package main

import (
	"errors"
	"fmt"
)

var ErrBlah = errors.New("blah")

func m() error { return ErrBlah }
func n() error { return ErrBlah }

func main() {
	fmt.Println("Hello, playground")

	a := errors.New("blah")
	b := errors.New("blah")

	fmt.Println(a == b)     // false
	fmt.Println(m() == n()) // true
}
```

### DIFF
--- a/cmd/appgw-ingress/errors.go
+++ b/cmd/appgw-ingress/errors.go
@@ -1,0 +1,12 @@
+// -------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// --------------------------------------------------------------------------------------------
+
+package main
+
+import "errors"
+
+var (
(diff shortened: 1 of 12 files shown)


cmd/appgw-ingress/errors.go
```diff
@@ -0,0 +1,10 @@
+// -------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// --------------------------------------------------------------------------------------------
+
```
akshaysngupta: can we move this to `errors` package ?

delqn: I don' think it's necessary at this point.

PR #412, opened 2019-07-24
Improve logging

delqn:
This PR slightly improves logging, by providing information on Namespaces and Blacklist when AGIC starts.

PR #418, opened 2019-07-25
brownfield: TargetPath.contains(TargetPath)

delqn:
In this PR:

 - Introducing new type `TargetPath`, which is a string -- a URL Path - `/x/*` for example
 - adding a `contains` function to inform whether path N is contained within path M

Example:
```yaml
apiVersion: appgw.ingress.k8s.io/v1
kind: AzureIngressProhibitedTarget
metadata:
  name: blacklist-all-targets
spec:
  paths:
  - /*
```

should blacklist `/abc` and all other paths.

(This here is required for https://github.com/Azure/application-gateway-kubernetes-ingress/pull/412/files#diff-09ce61e60cf8eae7f9ff52a875eafea7R40)

### DIFF
--- a/pkg/brownfield/targets_test.go
+++ b/pkg/brownfield/targets_test.go
@@ -26,7 +26,7 @@ var _ = Describe("Test blacklisting targets", func() {
 
 			// Targets /fox and /bar are in the blacklist
 			for _, path := range []string{fixtures.PathFox, fixtures.PathBar} {
-				expected.Path = path
+				expected.Path = TargetPath(path)
 				Expect(*blacklist).To(ContainElement(expected))
 			}
 			expected.Path = fixtures.PathFoo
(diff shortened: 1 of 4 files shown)


pkg/brownfield/targets_test.go
```diff
@@ -86,4 +86,37 @@ var _ = Describe("Test blacklisting targets", func() {
...
+			Expect(TargetPath("/*").contains("/")).To(BeTrue())
+			Expect(TargetPath("/*").contains("")).To(BeTrue())
+			Expect(TargetPath("/*").contains("/*")).To(BeTrue())
+
+			Expect(TargetPath("*").contains("/blah")).To(BeTrue())
+			Expect(TargetPath("*").contains("/")).To(BeTrue())
+			Expect(TargetPath("*").contains("")).To(BeTrue())
+			Expect(TargetPath("*").contains("/*")).To(BeTrue())
+
+			Expect(TargetPath("/").contains("/blah")).To(BeTrue())
```
akshaysngupta: should this be an exact match ?

akshaysngupta:
Great!

pkg/brownfield/targets_test.go
delqn: Let's do it!

PR #419, opened 2019-07-25
brownfield: Redirect Configurations

delqn:
Applying all the same patterns as we have done with the other App Gateway configuration resources - we now apply `AzureIngressProhibitedTarget` (blacklist) CRD to the `RedirectConfigurations`.

Additionally - adding helper to dump the config we obtained from App Gateway (at `glog.V(5)`)

### DIFF
--- a/pkg/brownfield/redirects.go
+++ b/pkg/brownfield/redirects.go
@@ -1,0 +1,90 @@
+// -------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// --------------------------------------------------------------------------------------------
+
+package brownfield
+
+import (
+	"strings"
+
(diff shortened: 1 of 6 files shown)


pkg/brownfield/redirects.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/utils"
+)
+
+type redirectName string
+type redirectsByName map[redirectName]n.ApplicationGatewayRedirectConfiguration
+
+// GetBlacklistedRedirects filters the given list of health probes to the list Probes that AGIC is allowed to manage.
+func (er ExistingResources) GetBlacklistedRedirects() ([]n.ApplicationGatewayRedirectConfiguration, []n.ApplicationGatewayRedirectConfiguration) {
+	blacklistedListeners := er.getBlacklistedListenersSet()
```
akshaysngupta: we need to start with GetBlacklistedRoutingRules

PR #422, opened 2019-07-25
chore: refactor url path map generation

akshaysngupta:
This PR aims to simiplify url path map creation.

### DIFF
--- a/pkg/appgw/requestroutingrules.go
+++ b/pkg/appgw/requestroutingrules.go
@@ -60,7 +60,7 @@ func (c *appGwConfigBuilder) RequestRoutingRules(cbCtx *ConfigBuilderContext) er
 }
 
 func (c *appGwConfigBuilder) getRules(cbCtx *ConfigBuilderContext) ([]n.ApplicationGatewayRequestRoutingRule, []n.ApplicationGatewayURLPathMap) {
-	httpListenersMap := c.groupListenersByListenerIdentifier(c.appGw.HTTPListeners)
+	httpListenersMap := c.groupListenersByListenerIdentifier(c.getListeners(cbCtx))
 	var pathMap []n.ApplicationGatewayURLPathMap
 	var requestRoutingRules []n.ApplicationGatewayRequestRoutingRule
 	for listenerID, urlPathMap := range c.getPathMaps(cbCtx) {
@@ -219,7 +219,6 @@ func (c *appGwConfigBuilder) getPathRules(cbCtx *ConfigBuilderContext, listenerI
 	pathRules := make([]n.ApplicationGatewayPathRule, 0)
 	for pathIdx := range rule.HTTP.Paths {
 		path := &rule.HTTP.Paths[pathIdx]
-
 		if len(path.Path) == 0 || path.Path == "/*" || path.Path == "/" {
 			continue
 		}
--- a/pkg/appgw/requestroutingrules.go
+++ b/pkg/appgw/requestroutingrules.go
@@ -59,131 +59,11 @@ func (c *appGwConfigBuilder) RequestRoutingRules(cbCtx *ConfigBuilderContext) er
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/appgw/requestroutingrules.go
```diff
@@ -59,14 +59,50 @@ func (c *appGwConfigBuilder) RequestRoutingRules(cbCtx *ConfigBuilderContext) er
...
+	defaultHTTPSettingsID := to.StringPtr(c.appGwIdentifier.httpSettingsID(defaultBackendHTTPSettingsName))
+	httpListenersMap := c.groupListenersByListenerIdentifier(c.appGw.HTTPListeners)
 	urlPathMaps := make(map[listenerIdentifier]*n.ApplicationGatewayURLPathMap)
-	backendPools := c.newBackendPoolMap(cbCtx)
-	_, backendHTTPSettingsMap, _, _ := c.getBackendsAndSettingsMap(cbCtx)
-	for _, ingress := range cbCtx.IngressList {
-		defaultAddressPoolID := c.appGwIdentifier.addressPoolID(defaultBackendAddressPoolName)
-		defaultHTTPSettingsID := c.appGwIdentifier.httpSettingsID(defaultBackendHTTPSettingsName)
+	for ingressIdx := range cbCtx.IngressList {
+		ingress := cbCtx.IngressList[ingressIdx]
```
rheaj99: Why do we change for `for _, ingress := range cbCtx.IngressList` to `for ingressIdx := range cbCtx.IngressList` if we only use `ingressIdx` to get the `ingress`?

pkg/appgw/requestroutingrules.go
```diff
@@ -59,131 +59,11 @@ func (c *appGwConfigBuilder) RequestRoutingRules(cbCtx *ConfigBuilderContext) er
...
-			},
-		}
-	}
-
-	return urlPathMaps
-}
-
 func (c *appGwConfigBuilder) getRules(cbCtx *ConfigBuilderContext) ([]n.ApplicationGatewayRequestRoutingRule, []n.ApplicationGatewayURLPathMap) {
-	httpListenersMap := c.groupListenersByListenerIdentifier(c.getListeners(cbCtx))
+	httpListenersMap := c.groupListenersByListenerIdentifier(c.appGw.HTTPListeners)
```
delqn: This is making the assumption that we have `c.appGw.HTTPListeners` already - but I don't think we have guarantees for that here.  If we keep this line - we would have an implicitly required order of execution, which is not a good thing.

I propose we keep `c.getListeners(cbCtx)` here and if performance is a concern - we memoize it.

delqn: Here's a tweak to alleviate the side-effects concerts (duplicate log lines) - https://github.com/Azure/application-gateway-kubernetes-ingress/pull/445/files

pkg/appgw/requestroutingrules.go
akshaysngupta: This because we using the rule pointer to get the backendID from the map. This requires the actual objects address rather than a copies address.

akshaysngupta: https://github.com/Azure/application-gateway-kubernetes-ingress/pull/422/files#diff-9f4ecc464102fb9e5a7063f9a97115a3R240

pkg/appgw/ingress_rules.go
```diff
@@ -16,49 +16,62 @@ import (
 // processIngressRules creates the sets of front end listeners and ports, and a map of azure config per listener for the given ingress.
 func (c *appGwConfigBuilder) processIngressRules(ingress *v1beta1.Ingress, env environment.EnvVariables) (map[int32]interface{}, map[listenerIdentifier]listenerAzConfig) {
 	frontendPorts := make(map[int32]interface{})
+	listeners := make(map[listenerIdentifier]listenerAzConfig)
+	for ruleIdx := range ingress.Spec.Rules {
+		rule := &ingress.Spec.Rules[ruleIdx]
```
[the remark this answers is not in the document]
akshaysngupta: https://github.com/Azure/application-gateway-kubernetes-ingress/pull/422/files#diff-9f4ecc464102fb9e5a7063f9a97115a3R240

PR #443, opened 2019-07-31
prune redirect and no TLS ingresses

akshaysngupta:
This PR allows to ignore ingresses that have ssl-redirect true but no TLS which is an valid configuration.
We cannot configure HTTPS for it and since, it was meant to have redirect for HTTP, we should nto configure HTTP for it.

### DIFF
--- a/pkg/events/types.go
+++ b/pkg/events/types.go
@@ -24,5 +24,9 @@ const (
 	// ReasonPortResolutionError is a reason for an event to be emitted.
 	ReasonPortResolutionError = "PortResolutionError"
 
+	// ReasonNoPrivateIPError is a reason for an event to be emitted.
 	ReasonNoPrivateIPError = "NoPrivateIP"
+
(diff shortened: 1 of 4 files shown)


pkg/events/types.go
```diff
@@ -24,5 +24,9 @@ const (
 	// ReasonPortResolutionError is a reason for an event to be emitted.
 	ReasonPortResolutionError = "PortResolutionError"
 
+	// ReasonNoPrivateIPError is a reason for an event to be emitted.
 	ReasonNoPrivateIPError = "NoPrivateIP"
+
+	// ReasonRedirectwithNoTLS is a reason for an event to be emitted.
+	ReasonRedirectWithNoTLS = "RedirectWithNoTLS"
```
delqn: This does not seem to be used anywhere... perhaps... just yet...

akshaysngupta: `ReasonRedirectWithNoTLS` is used in pruneRedirectWithNoTLS function when creating the event.

PR #446, opened 2019-07-31
Correct memoization of getListeners

delqn:
The previous implementation was incorrect. The `c.appGw.HTTPListeners` contained App Gwy config we fetched from ARM on the arrival of a new K8s event.
This PR introduces a new struct, which will be `nil` on the arrival of a new event when we instantiate a new `appGwConfigBuilder ` with `NewConfigBuilder()`

To provide more context on this change:
1. The loop that receives Kubernetes events and launches Process(Event) [here](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/pkg/worker/worker.go#L33)

2. for each event we do a GET on App Gwy's config [here](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/pkg/controller/process.go#L31)
    
3. we pass the original config appGw into the NewConfigBuilder [here](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/pkg/controller/process.go#L95)

4. then we assign it to c.appGw [here](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/pkg/appgw/configbuilder.go#L42)

In other words we would not expect c.appGw.HTTPListeners to be nil.

### DIFF
--- a/pkg/appgw/configbuilder.go
+++ b/pkg/appgw/configbuilder.go
@@ -27,11 +27,16 @@ type ConfigBuilder interface {
 	PostBuildValidate(cbCtx *ConfigBuilderContext) error
 }
 
+type memoization struct {
+	listeners *[]n.ApplicationGatewayHTTPListener
+}
+
 type appGwConfigBuilder struct {
 	k8sContext      *k8scontext.Context
 	appGwIdentifier Identifier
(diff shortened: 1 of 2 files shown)


akshaysngupta:
@rheaj99 Thank!

PR #447, opened 2019-07-31
fix: remove * from health probe path

akshaysngupta:
This PR removes * from the probe path. Probe path can have start when rule.path = `/xyz*`.
This causes the probe to fail.

### DIFF
--- a/pkg/appgw/health_probes.go
+++ b/pkg/appgw/health_probes.go
@@ -8,6 +8,7 @@ package appgw
 import (
 	"fmt"
 	"sort"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/appgw/health_probes.go
```diff
@@ -105,6 +106,9 @@ func (c *appGwConfigBuilder) generateHealthProbe(backendID backendIdentifier) *n
 		}
 	}
 
+	if probe.Path != nil {
+		probe.Path = to.StringPtr(strings.TrimRight(*probe.Path, "*"))
```
delqn: We should add a unit test to illustrate what's going on here...

akshaysngupta: Will address this!

PR #448, opened 2019-07-31
Istio Routing Rules Set Up

rheaj99:
This PR defines some functions to help with Istio URL path maps and sets up the basic scaffolding in the requestroutingrules.go file


pkg/appgw/requestroutingrules.go
```diff
@@ -129,6 +130,28 @@ func (c *appGwConfigBuilder) getPathMaps(cbCtx *ConfigBuilderContext) map[listen
 		}
 	}
 
+	for _, virtualService := range cbCtx.IstioVirtualServices {
```
delqn: @rheaj99 could you please wrap this with `if cbCtx.EnableIstioIntegration {` ?

PR #449, opened 2019-07-31
feat: backend protocol

akshaysngupta:
This PR adds backend protocol annotation to specify http/https.
Currently, we are only going to add this annotation without providing a way to specify trusted root cert.

### DIFF
--- a/pkg/annotations/ingress_annotations.go
+++ b/pkg/annotations/ingress_annotations.go
@@ -7,6 +7,7 @@ package annotations
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/knative/pkg/apis/istio/v1alpha3"
 	"k8s.io/api/extensions/v1beta1"
@@ -37,8 +38,11 @@ const (
 	// SslRedirectKey defines the key for defining with SSL redirect should be turned on for an HTTP endpoint.
 	SslRedirectKey = ApplicationGatewayPrefix + "/ssl-redirect"
 
-	// UsePrivateIP defines the key to determine whether to use private ip with the ingress.
+	// UsePrivateIPKey defines the key to determine whether to use private ip with the ingress.
 	UsePrivateIPKey = ApplicationGatewayPrefix + "/use-private-ip"
+
+	// BackendProtocolKey defines the key to determine whether to use private ip with the ingress.
+	BackendProtocolKey = ApplicationGatewayPrefix + "/backend-protocol"
 
 	// IngressClassKey defines the key of the annotation which needs to be set in order to specify
 	// that this is an ingress resource meant for the application gateway ingress controller.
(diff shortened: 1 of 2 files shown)


pkg/appgw/backendhttpsettings.go
```diff
@@ -274,5 +273,11 @@ func (c *appGwConfigBuilder) generateHTTPSettings(backendID backendIdentifier, p
 		httpSettings.RequestTimeout = to.Int32Ptr(reqTimeout)
 	}
 
+	if backendProtocol, err := annotations.BackendProtocol(backendID.Ingress); err == nil {
+		if backendProtocol == annotations.HTTPS {
+			httpSettings.Protocol = n.HTTPS
+		}
+	}
```
delqn: It seems to me that if I annotated my Ingress incorrectly, I would not be told so by the software. It would be good to `glog` the error here (and perhaps `continue`).

pkg/annotations/ingress_annotations.go
```diff
@@ -108,6 +129,23 @@ func UsePrivateIP(ing *v1beta1.Ingress) (bool, error) {
...
 
+// BackendProtocol provides value for protocol to be used with the backend
+func BackendProtocol(ing *v1beta1.Ingress) (ProtocolEnum, error) {
+	protocol, err := parseString(ing, BackendProtocolKey)
+	if err != nil {
+		return HTTP, err
+	}
+
+	protocolLowerCase := strings.ToLower(protocol)
+	for protocolEnum, mappedProtocol := range ProtocolEnumLookup {
```
delqn: It seems that you are using the map al a list of tuples. It's a fine approach. But since we have a map we could turn this from ON to O1 lookup.

What if we inversed the map above to: `map[string]ProtocolEnum{"http": HTTP, "https": HTTPS}`
then here we could do `protocolEnum, ok := ProtocolEnumLookup[protocolLowerCase]`

pkg/appgw/backendhttpsettings.go
```diff
@@ -7,13 +7,12 @@ package appgw
...
 import (
 	"fmt"
-	"sort"
-
 	n "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-12-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/golang/glog"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"sort"
```
delqn: It'd be nice to move this ⤒ to standard lib section

pkg/annotations/ingress_annotations.go
akshaysngupta: thanks!

pkg/appgw/backendhttpsettings.go
akshaysngupta: along with glog, it might help to add an event on ingress as well.

PR #454, opened 2019-08-01
ingressStatus: iningress list, parse resourceid, fixtures

akshaysngupta:
This is a small PR part of https://github.com/Azure/application-gateway-kubernetes-ingress/pull/335


pkg/appgw/types.go
```diff
@@ -8,7 +8,7 @@ package appgw
 import (
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/environment"
 	"github.com/knative/pkg/apis/istio/v1alpha3"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
delqn: You shouldn't need the `v1` alias here.

PR #466, opened 2019-08-07
add clusterid in AppGw Tags

akshaysngupta:
This PR adds a cluster resource id tag to Application gateway resource.
 This tag is derived from the provider ID of the node which is set by the cloud provider.
We expect the provider ID to of the format
`/subscriptions/subid/resourcegroup/MC_resgp_aksresname/providers/Microsoft.Compute/virtualmachines/vm`
which is converted to aks resourceId.
image

### DIFF
--- a/pkg/appgw/appgw_test.go
+++ b/pkg/appgw/appgw_test.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/annotations"
+	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/aztags"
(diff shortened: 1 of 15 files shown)


pkg/k8scontext/types.go
```diff
@@ -22,6 +22,7 @@ type InformerCollection struct {
 	AzureIngressProhibitedTarget cache.SharedInformer
 	IstioGateway                 cache.SharedIndexInformer
 	IstioVirtualService          cache.SharedIndexInformer
+	Nodes                        cache.SharedIndexInformer
```
delqn: We are adding informers for Nodes only to get cluster ID.
We need to fetch the cluster ID once; Adding these informers here has long-term impact on the performance of AKS.
What if we came up with a different way to get a unique identifier for the cluster?
We could do something as simple as using the API host.

PR #471, opened 2019-08-12
Check probe port is service port before using it

### LINKED ISSUE #470: Auto generated health probes do not work with kubernetes sidecar pattern
**Describe the bug**
Auto generated health probes do not work with kubernetes sidecar pattern.

Assuming we have a container exposing some service at some port, and does not have a health check endpoint. Thus we spin up another companion container, implementing custom logic, and provide a health check endpoint at another port. Since the health check port is for internal use, we do not expose it through service and ingress.

**To Reproduce**

`kubectl apply -f /dev/stdin << EOF`

```yml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: some-server
spec:
  selector:
    matchLabels:
      app: some-server
  template:
    metadata:
      labels:
        app: some-server
    spec:
      containers:
      - name: nginx
        image: nginx:1.17
        ports:
          - containerPort: 80
        livenessProbe:
          httpGet:
            path: /healthz
            port: 8080
      - name: healthz
        image: gcr.io/google-containers/exechealthz-amd64:v1.3.0
        args:
          - -cmd
          - "true"
        ports:
          - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: some-service
spec:
  selector:
    app: some-server
  ports:
  - port: 80
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: some-ingress
  annotations:
spec:
  rules:
  - host: appgw.lan
    http:
      paths:
      - path: /
        backend:
          serviceName: some-service
          servicePort: 80
```

```
EOF
```

**Ingress Controller details**

0.7.1 (ef0838a), self built.

pallxk:
This is an attempt to fix #470, which checks health probe port is exposed by related service.


akshaysngupta:
Closing in favor of https://github.com/Azure/application-gateway-kubernetes-ingress/pull/482

PR #478, opened 2019-08-15
brownfield: Adding more brownfield tests

delqn:
This PR adds more tests to `pkg/brownfield`

Redirects definitely needs more work. Would like to merge the fixtures PR (https://github.com/Azure/application-gateway-kubernetes-ingress/pull/476) to avoid merge conflicts (or semantic merge conflicts)


pkg/brownfield/http_settings_test.go
```diff
@@ -6,13 +6,28 @@
...
+
+	n "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-12-01/network"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	ptv1 "github.com/Azure/application-gateway-kubernetes-ingress/pkg/apis/azureingressprohibitedtarget/v1"
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/tests/fixtures"
 )
 
+type mockLogger struct {
```
akshaysngupta: is this supposed to be in a different file ?

delqn: It would actually make sense to move it somewhere in the pkg/tests namespace

delqn: Moved to `pkg/tests/mocks`

akshaysngupta:
I think redirect tests can be improved.

PR #500, opened 2019-08-30
Adding a new ConfigBuilder test - wrapping all features in one test

delqn:
This PR adds a test, which is similar in nature to what we already have in `appgw_test.go`, but it actually gives you a visual of the entire JSON blob that was generated.

I need this, as I'm working on a bug, where ARM detects a missing definition of a reference.
This is not not caught by the smaller unit tests, since all components work correctly, but the final JSON has inconsistencies.

The bug manifests itself like this:
```json
I0829 20:22:17.158220    2882 process.go:130] BEGIN AppGateway deployment
I0829 20:22:17.614027    2882 utils.go:50] Saved App Gateway config to /tmp/app-gateway-config-1567135337612434400.json442945145
F0829 20:22:17.614088    2882 process.go:144] Failed applying App Gwy configuration: network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidResourceReference" Message="Resource /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Network/applicationGateways/applicationgatewayxxxx/redirectConfigurations/sslr-fl-ws.mis.li-443 referenced by resource /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Network/applicationGateways/xxxx/urlPathMaps/url-ws.mis.li-80 was not found. Please make sure that the referenced resource exists, and that both resources are in the same region." Details=[] -- {
```
We had fixed this issue, but it reappeared with the latest refactor of request routing rules. This tells me that we need a different approach to testing -- we need tests that wrap multiple features into one test and look at the output holistically.
(This test here is a false negative - will be addressed with another PR)

I hesitate to call this here an integration test, but it closer to that than the unit-tests we have added to this repo already.


pkg/appgw/configbuilder_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+			jsonBlob, err = json.MarshalIndent(into, "--", "    ")
+			Expect(err).ToNot(HaveOccurred())
+
+			jsonTxt := string(jsonBlob)
+
+			expected := `{
+--    "properties": {
+--        "backendAddressPools": [
+--            {
```
akshaysngupta: I think this backend pool is missing a backend pool. Maybe something is wrong in the setup ?

akshaysngupta:
@delqn In processIngressRule, The line `hasTLS = cert != nil` => `hasTLS = false` as cert (secret doesn't exists) is missing from the cert store.
This causes it to return two listener config both with `HTTP` protocol instead of having `HTTP` and `HTTPS`.

delqn:
@akshaysngupta your assessment aligns with mine. I'm a) testing for this here in this PR and b) adding this back w/ https://github.com/Azure/application-gateway-kubernetes-ingress/pull/501

PR #501, opened 2019-08-30
ARM reports a missing definition of a referenced resource

delqn:
This PR fixes the bug described and false-negative tested in https://github.com/Azure/application-gateway-kubernetes-ingress/pull/500

The symptom of the bug - ARM detects a missing redirect; App Gateway config references a redirect, but the actual redirect is not defined in the JSON blob we sent to ARM.

This is not not caught by the smaller unit tests, since all components work correctly, but the final JSON is inconsistent.

The bug manifests itself like this:
```json
I0829 20:22:17.158220    2882 process.go:130] BEGIN AppGateway deployment
I0829 20:22:17.614027    2882 utils.go:50] Saved App Gateway config to /tmp/app-gateway-config-1567135337612434400.json442945145
F0829 20:22:17.614088    2882 process.go:144] Failed applying App Gwy configuration: network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidResourceReference" Message="Resource /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Network/applicationGateways/applicationgatewayxxxx/redirectConfigurations/sslr-fl-ws.mis.li-443 referenced by resource /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Network/applicationGateways/xxxx/urlPathMaps/url-ws.mis.li-80 was not found. Please make sure that the referenced resource exists, and that both resources are in the same region." Details=[] -- {
```
We had fixed this issue, but it reappeared with the latest refactor of request routing rules. To prevent this from happening, I am adding tests that a) make it obvious what the final JSON is and b) wrap multiple features into one test and look at the output holistically.

This test changed in this PR was introduced deliberately as a false negative and is now fixed to show that the bug existed and is addressed with this change.

### DIFF
--- a/docs/examples/guestbook/ing-guestbook-other.yaml
+++ b/docs/examples/guestbook/ing-guestbook-other.yaml
@@ -4,6 +4,7 @@ metadata:
   name: guestbook
   annotations:
     kubernetes.io/ingress.class: azure/application-gateway
+    appgw.ingress.kubernetes.io/backend-path-prefix: /
 spec:
   rules:
   - http:
--- a/pkg/appgw/configbuilder.go
+++ b/pkg/appgw/configbuilder.go
@@ -43,6 +43,7 @@ type memoization struct {
 	pools                        *[]n.ApplicationGatewayBackendAddressPool
 	certs                        *[]n.ApplicationGatewaySslCertificate
 	secretToCert                 *map[secretIdentifier]*string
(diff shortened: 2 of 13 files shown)


delqn:
Tagging this with do-not-merge and wip since it will cause a semantic merge conflict with https://github.com/Azure/application-gateway-kubernetes-ingress/pull/500 -- both PRs could pass independently, but together they would cause a failure.

PR #518, opened 2019-09-05
Simplify grouping of Frontend Listeners

delqn:
Since `lookupFrontendPortByID` is trivial and we are not individually testing it - moving the loop (and converting it into a map) into `groupListenersByListenerIdentifier`

Renaming `groupListenersByListenerIdentifier` to `groupListeners`


delqn:
Closing in lieu of https://github.com/Azure/application-gateway-kubernetes-ingress/pull/525

PR #522, opened 2019-09-05
Add suffixes to front end poirts based on public/private IP usage

delqn:
This PR adds a string suffix `{"priv", "pub"}` to:
  - Listener Names
  - Request Routing Rule Names

This is necessary to avoid conflicts in JSON submitted to ARM when we use both Public and Private IP addresses (on two separate Listeners)

Error this resolves:
```
F0905 15:55:33.919047   20810 process.go:147] Failed applying App Gwy configuration: network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequestFormat" Message="Cannot parse the request." Details=[{"code":"DuplicateResourceName","message":"Resource /subscriptions//resourceGroups//providers/Microsoft.Network/applicationGateways/ has two child resources with the same name (rr-80)."}] -- {
```


delqn:
Closing in lieu of https://github.com/Azure/application-gateway-kubernetes-ingress/pull/525

PR #523, opened 2019-09-05
Using both Public & Private IP on App Gateway

delqn:


delqn:
Moved important pieces to https://github.com/Azure/application-gateway-kubernetes-ingress/pull/525

PR #527, opened 2019-09-06
Drain the Work channel; Use the last event only

delqn:
This PR introduces a new function - `drainChan()`

The goal is to always work on the very last event within our buffered channel.
This is acceptable, since the Ingress Controller always queries for the Ingress, Service, Endpoints etc. and does not need the entire chain of events.

This will prevent overwhelming ARM with too many consecutive PUTs when the channel fills up quicker than ARM can apply config.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -62,7 +62,7 @@ func (c *AppGwIngressController) Start(envVariables environment.EnvVariables) er
 	}
 
 	// Starts Worker processing events from k8sContext
-	go c.worker.Run(c.k8sContext.Work, c.k8sContext.LastSync, c.stopChannel)
+	go c.worker.Run(c.k8sContext.Work, c.stopChannel)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


akshaysngupta:
@delqn do we need to drain even after taking in the ladt sync time?
Also, we need to solve the scarcity problem. If too many events come in, we may need to perform a force `process` call.

delqn:
@akshaysngupta I started this PR to analyzed how well the `last sync time` method works.

Here are some log lines showing that it does not quite achieve what we anticipated it to:
```
I0909 09:31:19.090552    5350 worker.go:20] Draining 0 events from work channel
I0909 09:31:19.241712    5350 worker.go:20] Draining 25 events from work channel
I0909 09:31:31.987085    5519 worker.go:20] Draining 85 events from work channel
I0909 09:33:09.054996    6275 worker.go:20] Draining 0 events from work channel
I0909 09:35:01.281743    6275 worker.go:20] Draining 942 events from work channel
```

Note that the draining happens after the timestamp comparison:
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/799964866394568781c3af88c5c43ca34a67f70e/pkg/worker/worker.go#L46-L51

PR #537, opened 2019-09-17
metrics: prometheus metric store

akshaysngupta:
This PR introduces a metric store targetting prometheus.


pkg/controller/process.go
```diff
@@ -153,7 +154,17 @@ func (c AppGwIngressController) Process(event events.Event) error {
...
 
 	// We keep this at log level 1 to show some heartbeat in the logs. Without this it is way too quiet.
-	glog.V(1).Infof("Applied App Gateway config in %+v", time.Now().Sub(deploymentStart).String())
+	duration := time.Now().Sub(deploymentStart)
+	glog.V(1).Infof("Applied App Gateway config in %+v", duration.String())
+	c.metricStore.UpdateMetric(metricstore.LatencyMetric, duration.Seconds())
+
+	// Publish metrics
+	for _, metric := range c.metricStore.GetMetrics() {
+		glog.V(3).Infof(metric.String())
```
delqn: 3 is supposed to be the quiet verbosity

pkg/environment/environment.go
```diff
@@ -6,10 +6,11 @@
 package environment
 
 import (
-	"github.com/pkg/errors"
 	"os"
 	"regexp"
 
+	"github.com/pkg/errors"
```
delqn: this should be `"errors"`

PR #538, opened 2019-09-19
events: plumb agic name/namespace & get agic pod

akshaysngupta:
This PR adds two env variable for AGIC Pod name/namespace in the deployment.yml file and adds a get pod function for AGIC pod. This will used later to populate events.


pkg/environment/environment.go
```diff
@@ -6,10 +6,11 @@
 package environment
 
 import (
-	"github.com/pkg/errors"
 	"os"
 	"regexp"
 
+	"github.com/pkg/errors"
```
delqn: this is just `errors`

pkg/k8scontext/context.go
```diff
@@ -171,6 +171,16 @@ func (c *Context) Run(stopChannel chan struct{}, omitCRDs bool, envVariables env
 	return nil
 }
 
+// GetAGICPod returns the pod with specified name and namespace
+func (c *Context) GetAGICPod(envVariables environment.EnvVariables) (*v1.Pod, error) {
+	pod, err := c.kubeClient.CoreV1().Pods(envVariables.AGICPodNamespace).Get(envVariables.AGICPodName, metav1.GetOptions{})
```
delqn: Why don't we move this one line somewhere above this: https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/cmd/appgw-ingress/main.go#L128
And there will be no need then to create this new exported method on the K8s Context struct.

pkg/controller/controller.go
```diff
@@ -30,6 +31,8 @@ type AppGwIngressController struct {
 
 	recorder record.EventRecorder
 
+	agicPod *v1.Pod
```
delqn: Should this be a pointer? If it is we have to deal with the `nil` case.
Is there a situation where we would not be able to retrieve the pod from the pod itself?  Is this a situation, in which we should be allowed to continue running the ingress controller?  It seems to me this will be a situation in which we want to halt and ask for correction of some kind.

pkg/controller/controller.go
```diff
@@ -61,6 +64,12 @@ func (c *AppGwIngressController) Start(envVariables environment.EnvVariables) er
 		return err
 	}
 
+	if pod, err := c.k8sContext.GetAGICPod(envVariables); err == nil {
```
delqn: This entire block should be combined with the implementation of the `GetAGICPod` and moved as a one-liner somewhere above this: https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/cmd/appgw-ingress/main.go#L128

This should happen once only during the lifetime of the pod, and it makes sense to put it near the initialization of the Controller.  We could pass the Pod identity into `NewAppGwIngressController`

pkg/controller/controller.go
akshaysngupta: yes, running on local machine may not have a POD.

pkg/controller/controller.go
akshaysngupta: sure

PR #539, opened 2019-09-19
httpserver: refactor health server to be more generic

akshaysngupta:
This PR makes way for adding a route for metrics in the http server already used by Health Probes.
In the following PR, we will add metric store which will provide a handler for "/metrics" path.


pkg/environment/environment.go
```diff
@@ -6,10 +6,11 @@
 package environment
 
 import (
-	"github.com/pkg/errors"
 	"os"
 	"regexp"
 
+	"github.com/pkg/errors"
```
delqn: This is just `errors`

Issue #594, opened 2019-10-07
docs rendering is wonky


sblack4:
**Describe the bug**
The yaml in bash toward the bottom of this document was being rendered incorrectly on the docs site. I changed the style to be in line with [your websockets page](https://azure.github.io/application-gateway-kubernetes-ingress/how-tos/websockets/) which renders it correctly. 

![image](https://user-images.githubusercontent.com/15880760/66355820-ab1fbb00-e936-11e9-82bf-631c3bd0c903.png)

Also the indentation under [Install AAD Pod Identity](https://azure.github.io/application-gateway-kubernetes-ingress/setup/install-new/#install-aad-pod-identity) had the bulleted list being rendered as a code block. 

![image](https://user-images.githubusercontent.com/15880760/66355904-efab5680-e936-11e9-9ae3-f2be0e83d097.png)
**To Reproduce**
Visit https://azure.github.io/application-gateway-kubernetes-ingress/setup/install-new/

**Ingress Controller details**
NA

sblack4:
It looks like my PR solved the indentation issue but not the `yaml` formatting issue:
![image](https://user-images.githubusercontent.com/15880760/66486938-841dd200-ea79-11e9-800d-752499f00ac0.png)

akshaysngupta:
There was a whitespace before the yaml code block tag.

PR #595, opened 2019-10-07
fix GH markdown causing rendering bug

sblack4:
#594

The yaml in bash toward the bottom of this document was being rendered incorrectly on the docs site. I changed the style to be in line with [your websockets page](https://azure.github.io/application-gateway-kubernetes-ingress/how-tos/websockets/) which renders it correctly. 

image

Also the indentation under [Install AAD Pod Identity](https://azure.github.io/application-gateway-kubernetes-ingress/setup/install-new/#install-aad-pod-identity) had the bulleted list being rendered as a code block. 

image

### DIFF
--- a/docs/setup/install-new.md
+++ b/docs/setup/install-new.md
@@ -175,7 +175,7 @@ Now that we have App Gateway, AKS, and AGIC installed we can install a sample ap
 via [Azure Cloud Shell](https://shell.azure.com/):
 
  ```yaml
- cat <<EOF | kubectl apply -f -
+cat <<EOF | kubectl apply -f -
 apiVersion: v1
 kind: Pod
 metadata:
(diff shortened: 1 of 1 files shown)


docs/setup/install-new.md
```diff
@@ -174,8 +174,9 @@ Kubernetes. We will leverage it to install the `application-gateway-kubernetes-i
 Now that we have App Gateway, AKS, and AGIC installed we can install a sample app
 via [Azure Cloud Shell](https://shell.azure.com/):
 
- ```bash
```
akshaysngupta: can you revert this ?

sblack4: Did it not fix the rendering? If you look at the picture none of that is being rendered as a code block

sblack4: But I can revert it no problem if need be

akshaysngupta: as in this part, `cat <<EOF | kubectl apply -f -` and `EOF` part.
Formatting change above looks good.

sblack4: done 😄

PR #665, opened 2019-11-14
docs: example for continuous deployment

akshaysngupta:


docs/how-tos/continuous-deployment.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Create a service principal to use with Azure Pipelines. This service principal will have owner access to current subscription. Access will be used to perform role assigement for AGIC identity in the pipeline.
+
+    ```bash
+    az ad sp create-for-rbac -n azure-pipeline-cd --role owner
+    # Copy the AppId and Password. We will use these in the next step.
+    ```
+
+1. Now, create a [new service connection](https://docs.microsoft.com/en-us/azure/devops/pipelines/library/service-endpoints?view=azure-devops&tabs=yaml#create-a-service-connection) in Azure Devops. Select "*use the full version of the service connection dialog*" option so that you can provide the newly created service principal.
+![Add Service Connection page](../images/pipeline-service-connection.png)
```
jaishals: Size is too big in the readme page. May be reduce the size

docs/how-tos/continuous-deployment.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Create a new Azure release pipeline
+
+We have prepared an [example release pipeline](./continuous-deployment-pipeline.json). This pipeline doesn the following:
```
jaishals: Nit: Typo

PR #701, opened 2020-01-03
annotations: add annotation for backend http setting hostname

### LINKED ISSUE #700: Allow to configure two different certs for public access and internal service
Hi,

This is a feature request.

After opening a support case and being followed by one colleague of yours, seems that is not posible to use two different certificates; one for the public URL accessed on WAF and the other one used by the pods, 
What I want to achieve is to have SSL channel between WAF and the pods, there's an article that points that is only necessary that the certificate is valid, but that's not accurate, after discussing the problem seems that it's a requirement to use the same hostname on public host as on the internal host… or have the public certificate to include names for the internal services.

E.G. 

CONFIG THAT WOULD WORK BUT NOT DESIRED:
"
spec:
 tls:
  - hosts:
    - $DNS_APP_FQDN_SVC
    secretName: $AKS_PUBLIC_CERT_NAME
 rules:
 - host: $DNS_APP_FQDN_SVC
   http:
    paths:
    - backend:
       serviceName: $APP_SVC
       servicePort: 443
"
The not desired config is because we don't want that internal cert used by the pods to match the public name.

DESIRED CONFIG THAT WON´T WORK:

"
spec:
 tls:
  - hosts:
    - $DNS_APP_FQDN_SVC
    secretName: $AKS_PUBLIC_CERT_NAME
 rules:
 - host: $DNS_APP_K8S_SVC
   http:
    paths:
    - backend:
       serviceName: $APP_SVC
       servicePort: 443
"
The public host name to be different from the healthprove for internal hosts. Please note that configuring WAF manually it would work (already tested).

If the explanation is a bit confusing I will be glad to contact any one that looks into this to give further details and explanations.

Thanks and kind regards,

akshaysngupta:
This PR adds `appgw.ingress.kubernetes.io/backend-hostname` to specify `BackendHttpSetting`'s hostname field.

Fixes #700


evmimagina:
Hi Akshay,

Wich version should I point to install and test the new feature?

Regards,

akshaysngupta:
Hi @evmimagina, This is not out yet. We will soon be publishing an rc3 which will have this feature.

evmimagina:
Hi Akshay,

I've tested the rc3 and it's not working, at least not out of the box... It seems that another adjustment needs to be done, when deploying using a personalised name for the back-end service, a back-end heath setting should be addapted as well;

image


"Pick host name from backend HTTP settings" instead of "no" and using a manual name (that matches the public name for the service), should be set on "yes", so it picks-up the name already configured on the HTTP settings section.

Regards,

akshaysngupta:
@evmimagina Thanks for pointing this out!

evmimagina:
Hi Akshay,

Thanks for the workaround, I will give it a try. Please keep me updated for a new RC so I can test the final solution.

Cheers,

evmimagina:
Hi @akshaysngupta , 

any updates on this?

Thanks,

evmimagina:
Dear @akshaysngupta ,

I've found the setting adjustment has been released on the 1.2.0-rc1. I've tested it and it works! thank you very much, you can close this request if you like.

Many thanks and best regards,

PR #704, opened 2020-01-07
retry: add a retry utility and replace `for` loops

akshaysngupta:
This PR introduces a retry utility to ease the retry process.
We, then, use the retry utility to replace azure auth retry and creation of Application Gateway.

For `getVnet`, we try for an extended period of time, which is 10 mins (60 tries with 10 sec gap) as master components have P95 of 10 mins.


cmd/appgw-ingress/main.go
```diff
@@ -161,14 +162,20 @@ func main() {
 
 	if err = azure.WaitForAzureAuth(azClient, maxAuthRetryCount, retryPause); err != nil {
```
jaishals: Can we please change the name  to some thing else. WaitForAppGateway ? CheckAppGwIsDeployed

jaishals: maxRetryCount .. there is nothing to do with Auth

cmd/appgw-ingress/main.go
```diff
@@ -161,14 +162,20 @@ func main() {
...
 		if err == azure.ErrAppGatewayNotFound && env.EnableDeployAppGateway {
-			if env.AppGwSubnetID != "" {
-				err = azClient.DeployGatewayWithSubnet(env.AppGwSubnetID)
-			} else if azContext != nil {
-				err = azClient.DeployGatewayWithVnet(azure.ResourceGroup(azContext.VNetResourceGroup), azure.ResourceName(azContext.VNetName), azure.ResourceName(env.AppGwSubnetName), env.AppGwSubnetPrefix)
-			}
+			// create application gateway with retries
+			err = utils.Retry(maxAuthRetryCount, retryPause,
+				func() (utils.Retriable, error) {
+					if env.AppGwSubnetID != "" {
```
jaishals: Can this be a separate function if its okay. ? Might be cleaner to understand the code here

cmd/appgw-ingress/main.go
```diff
@@ -161,14 +162,20 @@ func main() {
...
+						err = azClient.DeployGatewayWithSubnet(env.AppGwSubnetID)
+					} else if azContext != nil {
+						err = azClient.DeployGatewayWithVnet(azure.ResourceGroup(azContext.VNetResourceGroup), azure.ResourceName(azContext.VNetName), azure.ResourceName(env.AppGwSubnetName), env.AppGwSubnetPrefix)
+					}
+					glog.Error("Failed in deploying Application Gateway and dependencies: ", err)
```
jaishals: I think we should separate the deployment of VNet and AppGw. Retrying on failed appgw deployment doesnot make sense to me . 

We should separate it out so that deployment of appgw failing doesnot intervene with aks cluster coming up ? What do you think

pkg/appgw/configbuilder_test.go
```diff
@@ -685,8 +685,8 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Annotations: map[string]string{
-					annotations.IngressClassKey: annotations.ApplicationGatewayIngressClass,
-					annotations.SslRedirectKey:  "true",
+					annotations.IngressClassKey:      annotations.ApplicationGatewayIngressClass,
```
jaishals: Are these tabs ? or extra spaces ?

pkg/azure/client.go
```diff
@@ -102,8 +111,40 @@ func (az *azClient) SetAuthorizer(authorizer autorest.Authorizer) {
...
+			}
+
+			glog.Errorf("Failed fetching config for App Gateway instance. Will retry in %v. Error: %s", retryPause, err)
+			return utils.Retriable(true), ErrGetArmAuth
+		})
+
+	if err != nil && err != ErrAppGatewayNotFound {
+		glog.Errorf("Tried %d times to authenticate with ARM; Error: %s", retryCount, err)
+	}
+	return
```
jaishals: Do we not need to return values ?

pkg/appgw/configbuilder_test.go
akshaysngupta: these are spaces! I ran `go fmt ./...` on the entire project.

pkg/azure/client.go
akshaysngupta: This go's syntactic sugar. If a function uses named return parameters, then return can implicitly know what to return.
https://golang.org/doc/effective_go.html#named-results

Issue #705, opened 2020-01-09
appgw.usePrivateIP has no default value leading to errors


Morriz:
exactly as title says ;)

error:
```bash
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in ConfigMap.data.USE_PRIVATE_IP
```

akshaysngupta:
Thanks for reporting. 
Here is PR fixing it: https://github.com/Azure/application-gateway-kubernetes-ingress/pull/722

We will soon do a release carrying this fix.

PR #722, opened 2020-01-27
helm: specify a default value for usePrivateIP

akshaysngupta:


ferantivero:
hi @akshaysngupta just double checking this PR since today I faced the issue #705

as requested from https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/helm/ingress-azure/templates/configmap.yaml#L23 I think the default value require is for https://github.com/Azure/application-gateway-kubernetes-ingress/blob/c5997dbc5cac11b4aa7c096e43f1acd53f194ae9/helm/ingress-azure/values-template.yaml#L34

ferantivero:
@akshaysngupta just double checking this PR since today I faced the issue #705

 as requested from
 https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/helm/ingress-azure/templates/configmap.yaml#L23
 I think the value require is
 https://github.com/Azure/application-gateway-kubernetes-ingress/blob/c5997dbc5cac11b4aa7c096e43f1acd53f194ae9/helm/ingress-azure/values-template.yaml#L34

Heshma:
@akshaysngupta I'm still facing the same issue. 
The default value should be in appw snippet, as the value is referred as appw.usePrivateIP in configmap - https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/helm/ingress-azure/templates/configmap.yaml#L23, but usePrivateIP default value is added in kubernetes snippet in values.yaml. I think the fix should be revised.

PR #738, opened 2020-02-20
helm no longer uses tiller (v3)

joelharkes:
Helm no longer uses tiller, this seems to be no longer needed


akshaysngupta:
@joelharkes Thanks for contributing!

PR #742, opened 2020-02-23
probe: handle hostname extension

akshaysngupta:
This PR adds code to handle hostname extension annotation for the health probe host.
Probe host is derived from ingress/rule/host if readiness or liveness probe is not provided.

If there are more then 1 host name, then select the one that doesn't have a `*` or `?`. If none is found, then leave host field with the default value.

This is also responsible for intermittent failures in tests.


pkg/appgw/internaltypes.go
```diff
@@ -253,3 +257,15 @@ func (listenerID *listenerIdentifier) getHostNames() []string {
 
 	return hostnames
 }
+
+// getFirstHostNameWithouSpecialChars returns the first hostname which doesn't have special chars. To be used for probes.
+func (listenerID *listenerIdentifier) getFirstHostNameWithouSpecialChars() *string {
```
jaishals, line 262: Nit : Typo

pkg/appgw/internaltypes.go
```diff
@@ -39,6 +40,9 @@ const (
 const (
 	// MaxAllowedHostnames the maximum number of hostnames allowed for listener.
 	MaxAllowedHostnames int = 5
+
+	// WildcardSpecialCharacters are characters that are allowed for wildcard hostnames.
+	WildcardSpecialCharacters = "*?"
```
jaishals, line 45: Should we validate this in hostname ?

pkg/appgw/internaltypes.go
jaishals: Shall we call it GetHostNameForProbes

pkg/appgw/internaltypes.go
akshaysngupta: getHostNames function returns a list with the first hostname being the host in the ingress rule. So this should cover it. Is that what you mean ?

jaishals: No I meant, that do we do any validations when user commit the ingress, probably not. I was thinking we should allow users to commit ingress if extendedhostname has any other charactes which we dont support. 
But feel free to ignore as part of this review

akshaysngupta: This is what we will need. https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/

akshaysngupta:
This PR adds code to handle hostname extension annotation for the health probe host.
Probe host is derived from ingress/rule/host if readiness or liveness probe is not provided.

If there are more then 1 host name, then select the one that doesn't have a `*` or `?`. If none is found, then leave host field with the default value.

This is also responsible for intermittent failures in tests.

Issue #732, opened 2020-02-06
Question: How to support multiple AGIC with multiple AGW in a single AKS cluster using IngressClass


jsco2t:
Hi,

I have been digging through the documentation and I apologize if I have missed this. Here is the scenario I am trying to solve for:

- A single AKS Cluster
- Two App Gateways (configured with different routing/firewall rules)
- Two App Gateway Ingress Controllers

Basically (and this is over simplified). Say I have two kinds of workloads. Web Services and API Services. I want both workloads to run from the same cluster, but I need the Web "traffic" to be routed through one of the two App Gateways and the API "traffic" to be routed through the other App Gateway. In this scenario each App Gateway will have custom configuration in regards to it'security/firewall. 

As far as I can see - when you setup an AGIC (via Helm) you tell it which **single** AGW it should manage. Based on this - I was thinking that for my scenario I would need paired instances of AGIC's and AGW's

Web Services:
  AGW 1 <---> AGIC 1

API Services:
  AGW 2 <---> AGIC 2

I don't see any clear info in the documents about how to do this. Typically I would expect to see some sort of support (at least) for setting the `Ingress Class Name` for each Ingress Controller so that `Ingress Resources` could point to the correct controller (and by extension AGW):

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ingress-something
  labels:
    app: some-cool-api-trust-me
  annotations:
    kubernetes.io/ingress.class: azure/application-gateway <-----------------!!!
    appgw.ingress.kubernetes.io/backend-path-prefix: "/"
...
```

Any assistance you can provide would be greatly appreciated!

Thanks!

jsco2t:
I did see this issue:

https://github.com/Azure/application-gateway-kubernetes-ingress/issues/636

To further clarify the above -in my scenario both of the above described workloads live in the same `k8s namespace`.  I don't want to have to artificially segment my workloads into separate `k8s namespaces` unless that's the only way to make this work.

Even if the solution is segmenting `AGIC` by namespaces - it's still not fully clear to me how to deploy multiple of them to a single cluster (apologies if I am missing something obvious here).

Thanks!

akshaysngupta:
@jsco2t Unfortunately, At the moment, AGIC only supports partitioning on the basis of namespaces.

You are welcome to contribute code towards making ingress class configurable through helm config.

[...]

Issue #699, opened 2019-12-31
allow configuration of ingress annotation


Morriz:
The current annotation
```yaml
kubernetes.io/ingress.class: azure/application-gateway
```
does not play well with external-dns as it does not accept the label value (because of the slash in it). It should be possible to configure the label this controller listens to so we can make it valid:
```yaml
kubernetes.io/ingress.class: azure-application-gateway
```
Could you please make it so?

akshaysngupta:
@Morriz Interesting.
Would you be willing to take a stab at this ?

akshaysngupta:
Closing as it is covered by https://github.com/Azure/application-gateway-kubernetes-ingress/issues/732

PR #746, opened 2020-02-25
Proposed solution for issue #732

garvinmsft:


proposals/multiple-gateways-single-cluster.md
```diff
@@ -0,0 +1,70 @@
...
+     --ingress-suffix arbitrary-class
+```
+
+```yaml
+kind: Ingress
+metadata:
+  name: go-server-ingress-affinity
+  namespace: test-ag
+  annotations:
+    kubernetes.io/ingress.class: agic-arbitrary-class
```
akshaysngupta: currently, ingress.class is `azure/application-gateway` to AGIC.
So if it is set to `agic-<suffix>`, it will cause confusion as this changes the prefix as well.

garvinmsft: Thanks @akshaysngupta for your feedback. _agic-_ was just an example. _azure/application-gateway_ would work just as well. Would you like me to update the doc?

akshaysngupta: This won't fix the issue with External DNS in https://github.com/Azure/application-gateway-kubernetes-ingress/issues/699.

garvinmsft: To solve #699, we can move to a new prefix such as _agic-_ while maintaining backward compatibility with ingress resources still using _azure/application-gateway_. This would address all issues.

akshaysngupta: @garvinmsft I discussed with the team. Agreement is to allow "Allow any arbitrary string" for ingress.class when it is specified in the helm config.

garvinmsft: Sounds good. What should you like to do with this PR?

akshaysngupta: I will merge this in.

PR #769, opened 2020-03-20
Removed references to tiller

erjosito:
and set install command to helm3 syntax


akshaysngupta:
@erjosito Can you also add these changes to:
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/setup/install-new.md
and
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/setup/install-new-windows-cluster.md

erjosito:
@akshaysngupta done!

PR #775, opened 2020-03-25
reconcile: add opt-in reconciler event

akshaysngupta:
This PR provides the ability to reconcile Application Gateway based on a ticker event. Period for the ticker can be modified through helm/environment variable.

Since Application Gateway provides default for settings that were not set by AGIC, this leads to differences in the Generated config vs the config that is returned by ARM. To remove these differences, this PR also provides default values to those properties.


pkg/appgw/backendhttpsettings.go
```diff
@@ -205,8 +205,11 @@ func (c *appGwConfigBuilder) generateHTTPSettings(backendID backendIdentifier, p
 		Name: &httpSettingsName,
 		ID:   to.StringPtr(c.appGwIdentifier.HTTPSettingsID(httpSettingsName)),
 		ApplicationGatewayBackendHTTPSettingsPropertiesFormat: &n.ApplicationGatewayBackendHTTPSettingsPropertiesFormat{
-			Protocol: n.HTTP,
-			Port:     to.Int32Ptr(int32(port)),
+			Protocol:                       n.HTTP,
+			Port:                           to.Int32Ptr(int32(port)),
+			PickHostNameFromBackendAddress: to.BoolPtr(false), // setting to default
```
3quanfeng: different question here...
if we don't specify a field such as "PichHostName..." here, would appgw give it a default value as well?

pkg/environment/environment_test.go
```diff
@@ -159,5 +159,43 @@ var _ = Describe("Environment", func() {
...
+
+					ReconcilePeriodSeconds: "30",
+				}
+				Expect(ValidateEnv(env)).To(BeNil())
+
+				env.ReconcilePeriodSeconds = "300"
+				Expect(ValidateEnv(env)).To(BeNil())
+			})
+
+			It("should not error when input is out of range in RECONCILE_PERIOD_SECONDS", func() {
```
3quanfeng: "should error when input is out of range..."

pkg/worker/worker.go
```diff
@@ -29,8 +29,29 @@ func drainChan(ch chan events.Event, defaultEvent events.Event) events.Event {
...
+	}
+}
+
 // Run starts the worker which listens for events in eventChannel; stops when stopChannel is closed.
-func (w *Worker) Run(work chan events.Event, stopChannel chan struct{}) {
+func (w *Worker) Run(work chan events.Event, stopChannel chan struct{}, reconcilePeriodSeconds int) {
+
+	// initilize reconcilerEventTicker
+	if reconcilePeriodSeconds != -1 {
+		go reconcilerEventTicker(work, stopChannel, reconcilePeriodSeconds)
```
3quanfeng: It might be better to do the reconciler go routine  in controller.Start() given it's a totally different event, and the reconciler is just like a plugin, should be handled separately and the benefit is you can keep func signature clear as well as the test case.

pkg/controller/mutate_app_gateway.go
```diff
@@ -59,7 +59,16 @@ func (c AppGwIngressController) GetAppGw() (*n.ApplicationGateway, *appgw.Config
 }
 
 // MutateAppGateway applies App Gateway config.
-func (c AppGwIngressController) MutateAppGateway(appGw *n.ApplicationGateway, cbCtx *appgw.ConfigBuilderContext) error {
+func (c AppGwIngressController) MutateAppGateway(appGw *n.ApplicationGateway, cbCtx *appgw.ConfigBuilderContext, invokedForReconciliation bool) error {
```
3quanfeng: maybe we should do a complete separate MutateAppGateway for the reconcile case.
if somehow client wants to keep any fields of appgw config from the reconcile, the new generated from agic will just wipe them out, hence we need to document what fields agic will not wipe out, e.g. waf policy

pkg/controller/mutate_app_gateway.go
```diff
@@ -59,7 +59,16 @@ func (c AppGwIngressController) GetAppGw() (*n.ApplicationGateway, *appgw.Config
...
+	// if this is a reconciliation task, then before doing anything, compare the current expected state vs cached state of the gateway
+	if invokedForReconciliation {
+		glog.V(5).Info("[reconcile] triggered: ", invokedForReconciliation)
+		if c.configIsSame(appGw) {
```
3quanfeng: what if GetAppGw is different from cached config?

pkg/controller/controller.go
```diff
@@ -119,7 +119,8 @@ func (c *AppGwIngressController) ProcessEvent(event events.Event) error {
 		glog.Error("Error mutating AKS from k8s event. ", err)
 	}
 
-	if err := c.MutateAppGateway(appGw, cbCtx); err != nil {
+	invokedForReconciliation := event.Type == events.PeriodicReconcile
+	if err := c.MutateAppGateway(appGw, cbCtx, invokedForReconciliation); err != nil {
```
jaishals: Decision to whether mutate appgw due to any event should be togather. 
I think its better if we move out the logic of reconcile out of here

Should Provess Event should check for reconcileEvent, 
It should return false if any event has happened in time less than reconcile time. 
If it is decided to process that event, then enter ProcessEvent.

Process Event should not have logic of any specific event type

pkg/worker/worker_test.go
```diff
@@ -95,4 +95,30 @@ var _ = Describe("Worker Test", func() {
...
+			}
+			go worker.Run(work, stopChannel, "1")
+
+			// after 1 seconds, reconcilerTickerTask should add an event to the worker
+			processCalled := false
+			select {
+			case <-backChannel:
+				processCalled = true
+				break
+			case <-time.After(2 * time.Second):
```
3quanfeng: maybe 3 seconds for timeout?

pkg/worker/worker.go
akshaysngupta: good idea.

pkg/controller/mutate_app_gateway.go
akshaysngupta: Fields that are not overriden by AGIC will stay as is.

pkg/controller/mutate_app_gateway.go
akshaysngupta: If GetAppGw is different from cached config, then we proceed forward. We use the current gateway state and add back rules that AGIC expects. Since we start by using the current state of the app gateway, our new state has the properties that are part of the current state. Only rules part is overriden.

pkg/controller/controller.go
akshaysngupta: sure, I can move the reconcile logic to ShouldProcess.

PR #790, opened 2020-04-02
migrate: from extensions/v1beta1 to networking/v1beta1 package

akshaysngupta:


pkg/k8scontext/supported_apiversion.go
```diff
@@ -0,0 +1,73 @@
...
+
+	err := runtimeScheme.Convert(old, networkingIngress, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return networkingIngress, nil
+}
+
+func toIngress(obj interface{}) (*networkingv1beta1.Ingress, bool) {
```
3quanfeng, line 40: let's add one neg unit test case for toIngress?

PR #797, opened 2020-04-06
Apply waf policy to httpListener without enabling appgw.waf_listener

3quanfeng:
Fix to apply waf policy to listener without enable appgw.waf_listener to true


pkg/appgw/frontend_listeners.go
```diff
@@ -115,10 +114,14 @@ func (c *appGwConfigBuilder) getListenerConfigs(cbCtx *ConfigBuilderContext) map
...
 		glog.V(5).Infof("Processing Rules for Ingress: %s/%s", ingress.Namespace, ingress.Name)
+		policy, err := annotations.WAFPolicy(ingress)
+		if len(policy) > 0 {
+			glog.V(5).Infof("Found WAF policy: %s", policy)
+		}
 		azListenerConfigs := c.getListenersFromIngress(ingress, cbCtx.EnvVariables)
 		for listenerID, azConfig := range azListenerConfigs {
-			if cbCtx.EnvVariables.AttachWAFPolicyToListener {
-				attachFirewallPolicy(cbCtx, ingress, &azConfig)
+			if cbCtx.EnvVariables.AttachWAFPolicyToListener || (err == nil && policy != "") {
```
akshaysngupta, line 123: should `err == nil` be used here ?

3quanfeng: > should err == nil be used here ?

The tuple is returned by annotations.WAFPolicy(ingress), we could ignore the err from the return, but I don't see any difference given the intention here?

pkg/appgw/requestroutingrules_test.go
```diff
@@ -233,6 +233,84 @@ var _ = Describe("Test routing rules generations", func() {
...
+		pathMap := configBuilder.getPathMaps(cbCtx)
+
+		rule := &ingress.Spec.Rules[0]
+		listenerID := generateListenerID(ingress, rule, n.HTTP, nil, false)
+		It("has waf policy in pathRule", func() {
+			prs := pathMap[listenerID].ApplicationGatewayURLPathMapPropertiesFormat
+			for _, r := range *prs.PathRules {
+				Expect(r.FirewallPolicy.ID).To(Equal(to.StringPtr("/sub/waf")))
+			}
+		})
```
akshaysngupta, line 276: also, check that it is not applied to the listener.

pkg/appgw/requestroutingrules_test.go
```diff
@@ -233,6 +233,84 @@ var _ = Describe("Test routing rules generations", func() {
...
+
+		// !! Action !! -- will mutate pathMap struct
+		pathMap := configBuilder.getPathMaps(cbCtx)
+
+		rule := &ingress.Spec.Rules[0]
+		listenerID := generateListenerID(ingress, rule, n.HTTP, nil, false)
+		It("has no waf policy in pathRule", func() {
+			prs := pathMap[listenerID].ApplicationGatewayURLPathMapPropertiesFormat
+			Expect(prs.PathRules).To(BeNil())
+		})
```
akshaysngupta, line 311: check it is on listener ?

pkg/appgw/frontend_listeners.go
```diff
@@ -130,8 +128,13 @@ func (c *appGwConfigBuilder) getListenerConfigs(cbCtx *ConfigBuilderContext) map
 			// Default protocol
 			Protocol: n.HTTP,
 		}
-		if cbCtx.EnvVariables.AttachWAFPolicyToListener {
-			attachFirewallPolicy(cbCtx, nil, &listenerConfig)
+		// See if we have an ingress annotated with a Firewall Policy; Attach it to the listener
+		for _, ingress := range cbCtx.IngressList {
```
akshaysngupta, line 132: I don't think we need to assign it to the place holder listener.

pkg/appgw/frontend_listeners.go
```diff
@@ -140,6 +143,32 @@ func (c *appGwConfigBuilder) getListenerConfigs(cbCtx *ConfigBuilderContext) map
 	return allListeners
 }
 
+// applyWafPolicyToListener returns if we need to apply waf policy to a listener
+func (c *appGwConfigBuilder) applyWafPolicyToListener(cbCtx *ConfigBuilderContext) map[*v1beta1.Ingress]map[string]bool {
+	IngressApplyWafPolicytoListener := make(map[*v1beta1.Ingress]map[string]bool)
+	slash := "/"
```
akshaysngupta, line 149: make these as const ? later, we can use them everywhere else ?

pkg/appgw/ingress_rules.go
```diff
@@ -14,23 +14,46 @@ import (
...
+			}
 			listeners[k] = v
 		}
 	}
+
 	return listeners
 }
 
-func (c *appGwConfigBuilder) processIngressRule(rule *v1beta1.IngressRule, ingress *v1beta1.Ingress, env environment.EnvVariables) (map[Port]interface{}, map[listenerIdentifier]listenerAzConfig) {
+func (c *appGwConfigBuilder) processIngresswithWafPolicy(azConfig *listenerAzConfig, ingress *v1beta1.Ingress, env environment.EnvVariables) {
```
akshaysngupta, line 44: nit: `With`

pkg/appgw/ingress_rules.go
```diff
@@ -14,23 +14,46 @@ import (
...
+	policy, _ := annotations.WAFPolicy(ingress)
+	if policy != "" {
+		if env.AttachWAFPolicyToListener {
+			// logging to see if customer configures env.AttachWAFPolicyToListener or not
+			glog.V(5).Info("AttachWAFPolicyToListener is enabled")
+		}
+		glog.V(5).Infof("Found WAF policy: %s in annotation", policy)
```
akshaysngupta, line 51: glog.V(5).Infof("Attaching WAF policy: %s to listener <> ?", policy)
Do we need to add info about the listener as well ? Maybe the host name ?

pkg/appgw/ingress_rules.go
```diff
@@ -14,23 +14,46 @@ import (
...
 		rule := &ingress.Spec.Rules[ruleIdx]
 		if rule.HTTP == nil {
 			continue
 		}
 
-		_, ruleListeners := c.processIngressRule(rule, ingress, env)
+		_, ruleListeners := c.processIngressRuleWithTLS(rule, ingress, env)
+
 		for k, v := range ruleListeners {
+			if ingressWafPolicy != nil && ingressWafPolicy[ingress][rule.Host] {
```
akshaysngupta, line 34: we can evaluate the map inside the function

pkg/appgw/frontend_listeners_test.go
```diff
@@ -444,4 +444,72 @@ var _ = Describe("MutateAppGateway ingress rules and parse frontend listener con
...
+				},
+				EnvVariables:          envVariables,
+				DefaultAddressPoolID:  to.StringPtr("xx"),
+				DefaultHTTPSettingsID: to.StringPtr("yy"),
+			}
+
+			listeners, _ := cb.getListeners(cbCtx)
+
+			expectedListener443.FirewallPolicy = resourceRef(wafPolicyID)
+			Expect(*listeners).NotTo(ContainElement(expectedListener443))
```
akshaysngupta, line 467: Looking for `expectedListener443` might be error prone.

pkg/appgw/requestroutingrules_test.go
3quanfeng: > also, check that it is not applied to the listener.

It's covered in front_listener_test.go

pkg/appgw/frontend_listeners_test.go
```diff
@@ -444,4 +444,72 @@ var _ = Describe("MutateAppGateway ingress rules and parse frontend listener con
 			Expect(*listeners).To(ContainElement(expectedListener443))
 		})
 	})
+
```
akshaysngupta, line 447: we should also add a positive test ?

pkg/appgw/ingress_rules.go
3quanfeng: > nit: With
Updated

pkg/appgw/frontend_listeners_test.go
3quanfeng: > we should also add a positive test ?

The positive case is already covered in functional test regarding waf policy

pkg/appgw/ingress_rules.go
3quanfeng: > we can evaluate the map inside the function

good catch

PR #801, opened 2020-04-08
Add support for custom ingress class value

### LINKED ISSUE #732: Question: How to support multiple AGIC with multiple AGW in a single AKS cluster using IngressClass

garvinmsft:
This PR add support for making ingress class configurable through helm.
Here is the proposal document: https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/proposals/multiple-gateways-single-cluster.md

Fixes #732


pkg/annotations/ingress_annotations.go
```diff
@@ -59,9 +59,8 @@ const (
 	// that this is a gateway meant for the application gateway ingress controller.
 	IstioGatewayKey = "appgw.ingress.istio.io/v1alpha3"
 
-	// ApplicationGatewayIngressClass defines the value of the `IngressClassKey` and `IstioGatewayKey`
-	// annotations that will tell the ingress controller whether it should act on this ingress resource or not.
-	ApplicationGatewayIngressClass = "azure/application-gateway"
+	//DefApplicationGatewayIngressClass defines the default app gateway ingress value
+	DefApplicationGatewayIngressClass = "azure/application-gateway"
```
3quanfeng, line 63: Why don't we just name it "DefaultApplicationGatewayIngressClass"?

cmd/appgw-ingress/main.go
```diff
@@ -79,6 +80,12 @@ func main() {
 
 	env.Consolidate(cpConfig)
 
+	// adjust ingress class value if overridden by environment variable
+	if env.IngressClass != "" {
```
3quanfeng, line 84: Maybe we can do the assignment in utils.go

pkg/annotations/ingress_annotations_test.go
```diff
@@ -215,6 +220,22 @@ var _ = Describe("Test ingress annotation functions", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(actual).To(Equal(true))
 		})
+
+		It("returns true with correct annotation", func() {
+			ing.Annotations[IngressClassKey] = "agis5"
+			ApplicationGatewayIngressClass = "agis5"
+			actual, err := IsApplicationGatewayIngress(ing)
```
3quanfeng, line 227: I would suggest to update IsApplicationGatewayIngress to use DefaultApplicationGatewayIngressClass directly, benefit is you don't have to hack things outside to make the test work

pkg/annotations/ingress_annotations.go
garvinmsft: Name getting pretty long. Will change it.

cmd/appgw-ingress/main.go
garvinmsft: Thanks. I will look into this.

pkg/annotations/ingress_annotations_test.go
garvinmsft: I will look into this. I may have follow-up questions.

cmd/appgw-ingress/main.go
garvinmsft: @3quanfeng are you thinking something along the lines of:

`utils.SetIngressClass(env)`

pkg/annotations/ingress_annotations.go
akshaysngupta: this can also be `DefaultIngressClass`

cmd/appgw-ingress/main.go
3quanfeng: > @3quanfeng are you thinking something along the lines of:
> utils.SetIngressClass(env)

Yeah, something like getNamespacesToWatch(env.WatchNamespace), plus "pkg/annotations" is already imported in utils

pkg/annotations/ingress_annotations.go
garvinmsft: Even better!

pkg/annotations/ingress_annotations_test.go
garvinmsft: @3quanfeng can you share some additional thoughts in terms of hacking things? Which part would we be trying to avoid? The BeforeEach() block?

3quanfeng: Inside IsApplicationGatewayIngress, if you use DefaultIngressClass directly, then you don't need to care about ApplicationGatewayIngressClass in your test

garvinmsft: You mean like this? If yes, how would the ingress class override occur if not in this function?

```go

func IsApplicationGatewayIngress(ing *v1beta1.Ingress) (bool, error) {
	controllerName, err := parseString(ing, IngressClassKey)
	return controllerName == DefaultIngressClass, err
}
```

3quanfeng: Correct, the change honors the signature as well.

garvinmsft: This would negate the effect of this change:

https://github.com/garvinmsft/application-gateway-kubernetes-ingress/blob/92d9ac3a915f94e24a7f90883a99250164cd33c2/cmd/appgw-ingress/utils.go#L111

garvinmsft: The new logic intends to override the default. So that IsApplicationGatewayIngress returns true for the default or any arbitrary value set in IsApplicationGatewayIngress if overriden.
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/proposals/multiple-gateways-single-cluster.md

3quanfeng: I see, in this case, we could just set the default("azure/application-gateway") in environment.go.
then we don't need to make ANY changes in ingress_annotations.go, how does it sound?

3quanfeng: you can set the default in env.Consolidate()

garvinmsft: I think what you are getting at is to use 1 variable "ApplicationGatewayIngressClass" which contains the default but can be re-assigned in util. In that case, the tests won't cover the scenario where the ingress resource has a class different from the default but matches the new config value. 

What you are suggesting with env.Consolidate() can work but we would need to send env all the way to the handler and not rely on the package variable.

pkg/environment/environment.go
```diff
@@ -161,6 +165,7 @@ func GetEnv() EnvVariables {
 		AppGwResourceID:             os.Getenv(AppGwResourceIDVarName),
 		AppGwSubnetID:               os.Getenv(AppGwSubnetIDVarName),
 		AuthLocation:                os.Getenv(AuthLocationVarName),
+		IngressClass:				 os.Getenv(IngressClass),
```
akshaysngupta, line 168: nit: spacing

pkg/annotations/ingress_annotations_test.go
```diff
@@ -215,6 +220,22 @@ var _ = Describe("Test ingress annotation functions", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(actual).To(Equal(true))
 		})
+
+		It("returns true with correct annotation", func() {
+			ing.Annotations[IngressClassKey] = "agis5"
```
akshaysngupta, line 225: type: agis5 -> custom-class

PR #803, opened 2020-04-09
Add client code for azurebackendpool(azbp) crd, upgrade k8s.io/client-go to work with k8s 1.16.7

3quanfeng:
ccp: add generated client code for azurebackendpool crd


3quanfeng:
Break the PR into small prs

PR #808, opened 2020-04-14
Generate client code for new CRD AzureBackendPool

3quanfeng:
1. Added CRD AzureBackendPool yaml definition files as well as examples
2. Added api types and client code generated by code-generator for the new CRD
3. Add one method, GetBackendPool in context.go


crds/AzureBackendPool.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: azurebackendpools.appgw.ingress.azure.io
+spec:
+  group: appgw.ingress.azure.io
+  versions:
+    - name: v1
```
akshaysngupta, line 8: v1 -> v1alpha1 ?

crds/AzureBackendPool.yaml
```diff
@@ -0,0 +1,38 @@
...
+                        description: "Subresource ID of the backend pool of the Application gateway"
+                        type: string
+                      ipAddresses:
+                        description: "A list of Pod IP addresses"
+                        type: array
+                        items:
+                          type: string
+  scope: Cluster
+  names:
+    plural: azurebackendpools
```
akshaysngupta, line 34: should this be application gateway backend pool ? (so as not to confuse with SLB's backend pool)

crds/AzureBackendPool.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: azurebackendpools.appgw.ingress.azure.io
+spec:
+  group: appgw.ingress.azure.io
```
akshaysngupta, line 6: Just to bring to notice, That we cannot use `azureingressprohibitedtargets.appgw.ingress.k8s.io` which was used for ProhibitedTarget CRD as sufficx `k8s.io` has been locked down.
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190612-crd-group-protection.md

crds/AzureBackendPool.yaml
3quanfeng: > should this be applicationgatewaybackendpool ? (so as not to confuse with SLB's backend pool)
Better name :)

crds/AzureBackendPool.yaml
3quanfeng: > v1 -> v1alpha1 ?
Sounds good.
Here is some link I find regarding the versioning
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/

akshaysngupta: will check this out

crds/AzureBackendPool.yaml
```diff
@@ -0,0 +1,38 @@
...
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
```
jaishals, line 15: We need to add aproperty for the resource Id of application gateway ?

crds/AzureBackendPool.yaml
```diff
@@ -0,0 +1,38 @@
...
+              type: object
+              properties:
+                backendPoolInfo:
```
jaishals, line 18: We need to add updateId  or GenerationId to correlate what update we are talking about?

crds/AzureBackendPool.yaml
3quanfeng: Good point

pkg/k8scontext/context.go
```diff
@@ -199,6 +200,26 @@ func (c *Context) GetAGICPod(envVariables environment.EnvVariables) *v1.Pod {
...
+	azpool, err := c.crdClient.AzureapplicationgatewaybackendpoolsV1beta1().AzureApplicationGatewayBackendPools().Get(backendPoolName, metav1.GetOptions{})
+	if err != nil {
+		glog.Error("Error fetching Azure application gateway backend pool resource, Error: ", err)
+		return nil
+	}
+	return azpool
+}
+
+// GetProhibitedTarget returns prohibited target with specified name and namespace
+func (c *Context) GetProhibitedTarget(namespace string, targetName string) *prohibitedv1.AzureIngressProhibitedTarget {
```
akshaysngupta, line 214: Let's remove these functions in a later PR.

PR #810, opened 2020-04-17
doc: Correct the field used for identityPrincipal

vavane-ms:
Since it's a managed identity, the field `principalId` should be used.


akshaysngupta:
@vavane-ms Thanks a lot. LGTM!

PR #821, opened 2020-04-24
listener: use hostnames instead of hostname in httpListener

akshaysngupta:
This PR shifts to the usage of listener.hostnames field instead of listener.hostname.


functional_tests/duplicate_ports.json
```diff
@@ -92,38 +92,40 @@
...
                     },
                     "frontendPort": {
-                        "id": "/subscriptions/--subscription--/resourceGroups/--resource-group--/providers/Microsoft.Network/applicationGateways/--app-gw-name--/frontendPorts/fp-443"
+                        "id": "/subscriptions/--subscription--/resourceGroups/--resource-group--/providers/Microsoft.Network/applicationGateways/--app-gw-name--/frontendPorts/fp-80"
                     },
-                    "hostName": "foo.baz",
-                    "hostnames": [],
-                    "protocol": "Https",
-                    "requireServerNameIndication": false,
-                    "sslCertificate": {
```
akshaysngupta, line 108: just noticed

akshaysngupta: actually the ordering of the listeners changed here.

pkg/appgw/internaltypes.go
```diff
@@ -230,17 +229,12 @@ func defaultBackendAddressPool(appGWIdentifier Identifier) n.ApplicationGatewayB
 func defaultFrontendListenerIdentifier() listenerIdentifier {
 	return listenerIdentifier{
 		FrontendPort: Port(80),
-		HostName:     "",
 	}
 }
 
 func (listenerID *listenerIdentifier) setHostNames(hostnames []string) {
```
3quanfeng, line 235: we might need some change in generateListenerID() as well in case rule.Host exisits

pkg/appgw/appgw_test.go
```diff
@@ -671,11 +670,10 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
 						FrontendPort:            resourceRef(frontendPortID),
 						SslCertificate:          resourceRef(sslCert),
 						Protocol:                n.HTTPS,
-						HostName:                &domainName,
```
3quanfeng, line 674: Do we have the case that rule.Host is specified in ingress spec without hostname-extension annotation? in this case, httpListeners.hostName will be "host1", and httpListeners.hostNames will be ["host1"]?

pkg/appgw/internaltypes.go
akshaysngupta: That is covered here: https://github.com/Azure/application-gateway-kubernetes-ingress/blob/b03f7e953aad8fd7a2b7629cbc9762023c0a901e/pkg/appgw/configbuilder.go#L227

pkg/appgw/appgw_test.go
akshaysngupta: If only rule.Host is specified, we will still set the hostNames property in the HttpListener.
`hostNames = [ "host1" ]`

Basically, we will stop using the `hostName` altogether.

pkg/appgw/internaltypes.go
3quanfeng: > That is covered here: 
>   
>     
>       application-gateway-kubernetes-ingress/pkg/appgw/configbuilder.go
>     
>     
>          Line 227
>       in
>       b03f7e9
>     
>     
>     
>     
> 
>         
>           
>            if rule != nil && rule.Host != "" {

how about we set listenerID.HostName = "xx" here?

akshaysngupta: is there any benefit?

3quanfeng: > is there any benefit?

if we don't hostname for listener anymore, I think we should be good

pkg/appgw/frontend_listeners_test.go
```diff
@@ -145,10 +142,9 @@ var _ = Describe("MutateAppGateway ingress rules and parse frontend listener con
 				FrontendIPConfiguration:     resourceRef(tests.PublicIPID),
 				FrontendPort:                resourceRef(resPref + "frontendPorts/fp-443"),
 				Protocol:                    n.ApplicationGatewayProtocol("Https"),
-				HostName:                    to.StringPtr(tests.Host),
 				SslCertificate:              resourceRef(resPref + "sslCertificates/--namespace-----the-name-of-the-secret--"),
 				RequireServerNameIndication: to.BoolPtr(false),
-				Hostnames:                   &[]string{},
+				Hostnames:                   &[]string{tests.Host},
```
3quanfeng, line 147: we might need to cover annotation hostname-extension for your code change?

akshaysngupta: we already have tests this annotation.

Issue #823, opened 2020-04-27
AGIC helm install incompatible with newly-released aad-pod-identity 1.6.0


phealy:
**Describe the bug**
aad-pod-identity released version 1.6.0 on Friday the 24th, which now enforces case sensitivity with CRD resources. The helm chart for AGIC does not use the proper capitalization and thus if you install AGIC after installing aad-pod-identity 1.6.0 it doesn't work. This is happening with 1.2.0-rc1. MIC logs show the following:
```shell
E0427 00:13:26.222815       1 mic.go:899] Ignoring azure identity default/agic-azid-ingress-azure, error: Invalid resource id: "", must match /subscriptions/<subid>/resourcegroups/<resourcegroup>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>
```

https://github.com/Azure/aad-pod-identity#v160-breaking-change

If you install AGIC first, the upgrade/install of aad-pod-identity 1.6.0 upgrades the existing identity objects, so it does work. I have worked around the issue for now by installing aad-pod-identity 1.5.5 explicitly:

```
kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/v1.5.5/deploy/infra/deployment-rbac.yaml
```

**To Reproduce**
Follow the greenfield install instructions. The cluster created will not have authentication because of aad-pod-identity 1.6.0.

**Ingress Controller details**
```shell
azure@Azure:~$ k describe pod agic-ingress-azure-7676f67786-pl7t2
Name:           agic-ingress-azure-7676f67786-pl7t2
Namespace:      default
Priority:       0
Node:           aks-agentpool-35064155-vmss000001/10.0.0.65
Start Time:     Mon, 27 Apr 2020 00:05:27 +0000
Labels:         aadpodidbinding=agic-ingress-azure
                app=ingress-azure
                pod-template-hash=7676f67786
                release=agic
Annotations:    prometheus.io/port: 8123
                prometheus.io/scrape: true
Status:         Running
IP:             10.0.0.75
IPs:            <none>
Controlled By:  ReplicaSet/agic-ingress-azure-7676f67786
Containers:
  ingress-azure:
    Container ID:   docker://9464c39f71ed55266e6871fdb1c2b84872e2b8d1a90548c394911b369a7e4e5a
    Image:          mcr.microsoft.com/azure-application-gateway/kubernetes-ingress:1.2.0-rc1
[...]

PR #825, opened 2020-04-27
fix: update the casing for aad-pod-identity crds

akshaysngupta:
This PR fix the casing in the AAD Pod identity CRDs created when using user assigned identity with AGIC.
This is to address the breaking change introduced in the AAD POD identity.
https://github.com/Azure/aad-pod-identity#v160-breaking-change

We create the AzureIdentity and AzureIdentityBinding resources.

Will go out in `1.2.0-rc2`.

Fixes https://github.com/Azure/application-gateway-kubernetes-ingress/issues/823


MarcelT-NL:
Is there an eta for this? Thanks!

Issue #824, opened 2020-04-27
AGIC not update backends when deployment replica chnage


nfinkel:
Hi All,

My AGIC is not detect any change on endpoint list of my service and ingress.

i'm using AGIC helm install version 1.0.0 with aad-pod-identity version 1.5.5
when i update the replica count of my deployment i don't see any change on the AGIC log or configuration.
if i delete and recreate my ingress, the new/deleted PODs are added to the endpoint list.

**Deployemnt**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

**Service**
```
apiVersion: v1
kind: Service
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  ports:
  - name: http
    port: 80
  selector:
    app: nginx
```

**Ingress**
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: nginx
  annotations:
    kubernetes.io/ingress.class: "azure/application-gateway"
    appgw.ingress.kubernetes.io/backend-protocol: "http"
spec:
  backend:
    serviceName: nginx
    servicePort: http
```

after deploy all 3 components, i can see the correct configuration on the AG with single endpoint.
then i increase the deployment replica count and there is no change on AG and no logs on AGIC,
i can see the new PODs added to the service and ingress as endpoints.
`kubectl scale deployment nginx-deployment --replicas=3`
but if i delete and apply the ingress, all 3 pods are added to the AG.

Appreciate your help
Nir

3quanfeng:
Hi @nfinkel , Thanks for reporting the issue, we are evaluating the issue, instead, could you please use the alternative in your ingress configure?
`spec:
  rules:
  - http:
      paths:
      - backend:
          serviceName: nginx
          servicePort: 80`

nfinkel:
Hi @3quanfeng 

The workaround work without any problem.
please find the updated ingress configuration.

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: nginx
  annotations:
    kubernetes.io/ingress.class: "azure/application-gateway"
    appgw.ingress.kubernetes.io/backend-protocol: "http"
spec:
  rules:
    - http:
[...]

PR #826, opened 2020-04-28
AGIC to handle ingress defined with default backend

3quanfeng:
1. AGIC worker to process pod event when ingress without rules is defined(#824)
2. Apply waf policy to listener when ingress without rules is defined


pkg/k8scontext/context.go
```diff
@@ -574,13 +574,20 @@ func (c *Context) listServicesByPodSelector(pod *v1.Pod) []*v1.Service {
...
+			for _, rule := range ingress.Spec.Rules {
+				for _, path := range rule.HTTP.Paths {
+					// TODO(akshaysngupta) Use service ports
+					if path.Backend.ServiceName == service.Name {
+						return true
+					}
 				}
 			}
+		} else {
+			// single service ingress without rules
```
akshaysngupta, line 587: can you have both backend and rules in the same ingress ?

pkg/appgw/requestroutingrules_test.go
```diff
@@ -311,6 +311,42 @@ var _ = Describe("Test routing rules generations", func() {
...
+		_ = configBuilder.BackendAddressPools(cbCtx)
+		_ = configBuilder.Listeners(cbCtx)
+
+		// !! Action !! -- will mutate pathMap struct
+		pathMap := configBuilder.getPathMaps(cbCtx)
+
+		listenerID := generateListenerID(ingress, nil, n.HTTP, nil, false)
+		It("has no waf policy in pathRule", func() {
+			prs := pathMap[listenerID].ApplicationGatewayURLPathMapPropertiesFormat
+			for _, r := range *prs.PathRules {
```
akshaysngupta, line 344: 1) Can you also check that the urlPathMap has the correct value for backend pool and backend http setting ?
2) Can you also add another test for where ingress has both rules and backend ?

akshaysngupta:
Also, looking at https://github.com/Azure/application-gateway-kubernetes-ingress/blob/3b45daa6e57eba0bd4879a67dde77ade733ae601/pkg/appgw/ingress_rules.go#L21

It might not work as expected when ingress.TLS or SslCertificate annotation is provided.

pkg/appgw/requestroutingrules_test.go
3quanfeng: > Can you also check that the urlPathMap has the correct value for backend pool and backend http setting ?
> Can you also add another test for where ingress has both rules and backend ?
Functional test might be a better place to verify if the appgw config generated as expected

pkg/appgw/certificates.go
```diff
@@ -115,7 +115,7 @@ func (c *appGwConfigBuilder) newHostToSecretMap(ingress *v1beta1.Ingress) map[st
 		}
 
 		// default secret
-		if len(tls.Hosts) == 0 {
+		if len(tls.Hosts) == 0 || len(ingress.Spec.Rules) == 0 {
```
akshaysngupta, line 118: This should be independent of ingress.Rules. I think this check will break ingress with rules without host.

pkg/k8scontext/context.go
```diff
@@ -365,18 +365,21 @@ func (c *Context) ListAzureProhibitedTargets() []*prohibitedv1.AzureIngressProhi
...
 
 	if !exist {
-		glog.V(9).Infof("Service %s does not exist", serviceKey)
+		glog.V(3).Infof("Service %s does not exist", serviceKey)
 		return nil
 	}
 
 	service := serviceInterface.(*v1.Service)
+	if service != nil {
+		glog.V(3).Infof("find service %s/%s", service.Namespace, service.Name)
```
akshaysngupta, line 380: V(5) ?

nit: found

pkg/appgw/ingress_rules.go
```diff
@@ -18,31 +18,43 @@ func (c *appGwConfigBuilder) getListenersFromIngress(ingress *v1beta1.Ingress, e
 	listeners := make(map[listenerIdentifier]listenerAzConfig)
 
 	// if ingress has only backend configured
-	if ingress.Spec.Backend != nil && len(ingress.Spec.Rules) == 0 {
+	if ingress.Spec.Backend != nil && ingress.Spec.TLS == nil && len(ingress.Spec.Rules) == 0 {
```
akshaysngupta, line 21: what if WAF Policy is defined on the ingress ?

pkg/appgw/ingress_rules.go
```diff
@@ -18,31 +18,43 @@ func (c *appGwConfigBuilder) getListenersFromIngress(ingress *v1beta1.Ingress, e
...
+		glog.V(3).Infof("Only default backend defined for the ingress %s/%s", ingress.Namespace, ingress.Name)
 		return listeners
 	}
 
 	// process ingress rules with TLS and Waf policy
 	policy, _ := annotations.WAFPolicy(ingress)
+	if len(ingress.Spec.Rules) == 0 {
```
akshaysngupta, line 28: Can this check be `ingress.Spec.Backend != nil` ?

functional_tests/functional_test.go
```diff
@@ -611,6 +612,133 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
...
+					ServiceName: serviceNameB,
+					ServicePort: intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 80,
+					},
+				},
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					annotations.IngressClassKey: annotations.ApplicationGatewayIngressClass,
```
akshaysngupta, line 695: Let's add backend-protocol annotation and waf-policy annotation

pkg/appgw/certificates.go
3quanfeng: > This should be independent of ingress.Rules. I think this check will break ingress with rules without host.

this check only happens when TLS defined, no matter you have ingress hosts or not

pkg/appgw/ingress_rules.go
3quanfeng: > what if WAF Policy is defined on the ingress with backend ?

good point

pkg/appgw/certificates.go
akshaysngupta: yes, we can treat this as loose binding. We can treat them independently.

pkg/appgw/ingress_rules.go
3quanfeng: The line is checking when ingress is defined with default backend and TLS, in which case, TLS will be applied to the default backend

> Can this check be `ingress.Spec.Backend != nil` ?

The line is checking when ingress is defined with default backend and TLS, in which case, TLS will be applied to the default backend

functional_tests/functional_test.go
3quanfeng: good point, will do

pkg/appgw/ingress_rules.go
3quanfeng: > what if WAF Policy is defined on the ingress with backend ?

This is actually covered, check the last test case in front_listener_test.go

functional_tests/functional_test.go
```diff
@@ -611,6 +660,180 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
...
+			Spec: v1beta1.IngressSpec{
+				Rules: []v1beta1.IngressRule{
+					{
+						IngressRuleValue: v1beta1.IngressRuleValue{
+							HTTP: &v1beta1.HTTPIngressRuleValue{
+								Paths: []v1beta1.HTTPIngressPath{
+									{
+										Path: "/Z/",
+										Backend: v1beta1.IngressBackend{
+											ServiceName: serviceNameZ,
```
akshaysngupta, line 751: Just confirming that `serviceNameZ` is deliberately missing right ?

functional_tests/functional_test.go
```diff
@@ -611,6 +660,180 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
...
+				Annotations: map[string]string{
+					annotations.IngressClassKey: annotations.ApplicationGatewayIngressClass,
+				},
+				Namespace: tests.Namespace,
+				Name:      tests.Name,
+			},
+		}
+
+		// Make sure ingress default backend is used as default
+		// The no matched rule backend in this case will still be used in urlPathMaps
```
akshaysngupta, line 781: what is the meaning of **no matched rule** ?

functional_tests/functional_test.go
3quanfeng: > Just confirming that `serviceNameZ` is deliberately missing right ?

Yes, you are right! means no match from backend

functional_tests/functional_test.go
3quanfeng: > what is the meaning of **no matched rule** ?
it means no backend and service pair can be found, in this case, the rule path if defined will still be added to the pathRules

pkg/appgw/requestroutingrules.go
```diff
@@ -158,13 +164,32 @@ func (c *appGwConfigBuilder) getPathMaps(cbCtx *ConfigBuilderContext) map[listen
...
 
+		//noMatchedService := true
 		for ruleIdx := range ingress.Spec.Rules {
 			rule := &ingress.Spec.Rules[ruleIdx]
 			// skip no http rule
 			if rule.HTTP == nil {
 				continue
 			}
 
+			// when rule backend defined but not matched
```
akshaysngupta, line 175: can you remove the comments ?

pkg/appgw/requestroutingrules.go
```diff
@@ -130,10 +130,16 @@ func (c *appGwConfigBuilder) noRulesIngress(cbCtx *ConfigBuilderContext, ingress
...
+
+	// Thre are no Rules, create urlPathMap with no pathRules
 	if serviceBackendPair, exists := serviceBackendPairMap[backendID]; exists {
 		poolName := generateAddressPoolName(backendID.serviceFullName(), backendID.Backend.ServicePort.String(), serviceBackendPair.BackendPort)
 		defaultAddressPoolID := c.appGwIdentifier.AddressPoolID(poolName)
-		defaultHTTPSettingsID := c.appGwIdentifier.HTTPSettingsID(DefaultBackendHTTPSettingsName)
+		//create default HttpSettings from ingress default backend
+		defaultHTTPSettingsID := c.appGwIdentifier.HTTPSettingsID(*noRulesIngressHTTPSettings.Name)
+
 		listenerID := defaultFrontendListenerIdentifier()
```
akshaysngupta, line 143: I think a similar change would be needed for frontend listener also. That is if the front end listener is having hostname-extension. Then default listener might not work. We can address this in a separate PR as well.

pkg/appgw/ingress_rules.go
```diff
@@ -18,37 +18,53 @@ func (c *appGwConfigBuilder) getListenersFromIngress(ingress *v1beta1.Ingress, e
...
-				glog.V(3).Infof("Attach WAF policy: %s to listener: %s", policy, generateListenerName(k))
-				v.FirewallPolicy = policy
-			}
-			listeners[k] = v
+func (c *appGwConfigBuilder) collectListener(ingress *v1beta1.Ingress, rule *v1beta1.IngressRule, env environment.EnvVariables, wafPolicy string) map[listenerIdentifier]listenerAzConfig {
+	listeners := make(map[listenerIdentifier]listenerAzConfig)
+	_, ruleListeners := c.processIngressRuleWithTLS(rule, ingress, env)
+	apply := false
+	if wafPolicy != "" {
+		apply = c.applyToListener(rule)
```
akshaysngupta, line 49: applyToListener -> applyWAFPolicyToListener

pkg/appgw/ingress_rules.go
```diff
@@ -18,37 +18,53 @@ func (c *appGwConfigBuilder) getListenersFromIngress(ingress *v1beta1.Ingress, e
 	listeners := make(map[listenerIdentifier]listenerAzConfig)
 
 	// if ingress has only backend configured
-	if ingress.Spec.Backend != nil && len(ingress.Spec.Rules) == 0 {
+	if ingress.Spec.Backend != nil && ingress.Spec.TLS == nil && len(ingress.Spec.Rules) == 0 {
```
akshaysngupta, line 21: I think we should not exit this function when TLS is not specified. This will miss the case for hostname-extension.

Can we do the following ? 
Remove
`if ingress.Spec.Backend != nil && ingress.Spec.TLS == nil && len(ingress.Spec.Rules) == 0 {`
and `if len(ingress.Spec.Rules) == 0 {`

and replace with a similar logic:
```go
if ingress.Spec.Backend != nil {
  listeners = c.collectListener(ingress, nil, env, policy)
}
```

functional_tests/functional_test.go
3quanfeng: Updated the comment into "No paired backend in this case will still be in urlPathMaps"

pkg/appgw/ingress_rules.go
3quanfeng: Updated

pkg/appgw/ingress_rules.go
3quanfeng: > I think we should not exit this function when TLS is not specified. This will miss the case for hostname-extension.
> 
> Can we do the following ?
> Remove
> `if ingress.Spec.Backend != nil && ingress.Spec.TLS == nil && len(ingress.Spec.Rules) == 0 {`
> and `if len(ingress.Spec.Rules) == 0 {`
> 
> and replace with a similar logic:
> 
> ```go
> if ingress.Spec.Backend != nil {
>   listeners = c.collectListener(ingress, nil, env, policy)
> }
> ```

 `if ingress.Spec.Backend != nil && ingress.Spec.TLS == nil && len(ingress.Spec.Rules) == 0 ` can be removed

3quanfeng: added one more function test case to cover two ingresses, one with default backend defined, the other not.

PR #839, opened 2020-05-04
e2e: add tests for wildcard hostname

akshaysngupta:
This PR adds e2e tests for wildcard hostname.

This PR is dependent on https://github.com/Azure/application-gateway-kubernetes-ingress/pull/821

image


scripts/e2e/run-e2e.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ex
+set -x
```
3quanfeng, line 2: why do we change to -x here?

scripts/e2e/run-e2e.sh
```diff
@@ -36,4 +36,4 @@ InstallAGIC
 
 # run test
 go mod init
-go test -v --tags e2e ./... > testoutput.txt || { echo "go test returned non-zero"; cat testoutput.txt; exit 1; }
+go test -v -timeout 60m -tags e2e ./... > testoutput.txt || { echo "go test returned non-zero"; cat testoutput.txt; exit 1; }
```
3quanfeng, line 39: why 60mins here? maybe we should set timeout per test case?

akshaysngupta: lets tackle that seperately.

PR #852, opened 2020-05-12
Fixed value to AAD POD identity

larshaupthansen:
According to https://github.com/Azure/aad-pod-identity this should be the value of the identity created not the fullname (ingress-azure).

Or the guide should mention, that the created identity needs to be called "ingress-azure".

### DIFF
--- a/helm/ingress-azure/templates/deployment.yaml
+++ b/helm/ingress-azure/templates/deployment.yaml
@@ -20,7 +20,7 @@ spec:
         release: {{ .Release.Name }}
         {{- if .Values.armAuth }}
         {{- if eq .Values.armAuth.type "aadPodIdentity"}}
-        aadpodidbinding: {{ template "application-gateway-kubernetes-ingress.fullname" . }}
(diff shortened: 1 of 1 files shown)


akshaysngupta:
@larshaupthansen `aadpodidbinding` is a label selector. It should match the selector criteria specified in the AzureIdentityBinding CRD. There is no requirement for this to be equal to `clientID`.
https://github.com/Azure/aad-pod-identity#6-deployment-and-validation
You can read the first line in the above link.
It says, "For a pod to match an identity binding, it needs a label with the key aadpodidbinding whose value is that of the selector: field in the AzureIdentityBinding."

larshaupthansen:
Ah yes, that is indeed the case and the helm chart creates the corresponding AzureIdentityBinding and AzureIdentity binding. 

I must look elsewhere to figure out why I got unauthorized when using the helm script as it is (it worked as soon as I did this change).

Both the AzureIdentity and the AzureIdentityBinding created the Helm chart and by me (using the scripts in https://github.com/Azure/aad-pod-identity ) look the same.

Issue #851, opened 2020-05-12
AGIC with HTTPS Not Fully Creating 


adamcarter81:
I am using AGIC with Istio which is fine over HTTP, but now I want to implement HTTPS to run alongside HTTP.
I have configured TLS passthrough in my Istio service with relevant wildcard certificate, I can test locally by forwarding a port and modifying my hosts file to point the domain to localhost - this works fine with no cert errors.
The issue is when I deploy Agic with what I think is a correct configuration - the front end listener and rules are not created
The backend pool, health probe and http settings are created OK

**To Reproduce**
`apiVersion: v1
kind: Service
metadata:
  name: agic-ingress-https
spec:
  selector:
    #app: istio-ingressgateway.istio-system.svc.cluster.local
    #app: istio-ingressgateway.istio-system
    app: istio-ingressgateway
    #app: azure-vote-front
    #host: istio-ingressgateway
  ports:
  - port: 443
    protocol: TCP
    name: https-443
    targetPort: 443

---

apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  #name: istio-ingressgateway-voting
  name: agic-ingress-https
  annotations:
    kubernetes.io/ingress.class: azure/application-gateway
    appgw.ingress.kubernetes.io/backend-hostname: "appgw.domain.com"
    appgw.ingress.kubernetes.io/cookie-based-affinity: "true"
    appgw.ingress.kubernetes.io/backend-protocol: "https"
    appgw.ingress.kubernetes.io/backend-path-prefix: "/"
    appgw.ingress.kubernetes.io/appgw-trusted-root-certificate: "istio-system-wildcard-domain.com"
spec:
  tls:
    - hosts:
      - server.domain.com
      secretName: wildcard-domain-com
  rules:
  - host: server.domain.com
  - http:
      paths:
      - path: /
        backend:
          serviceName: agic-ingress-https
          servicePort: 443`

**Ingress Controller details**
* Output of `kubectl describe pod <ingress controller`> .
`Name:           ingress-azure-7c85c7cfb6-k9qw2
Namespace:      default
Priority:       0
Node:           aks-linux-35064155-vmss000001/15.0.0.35
Start Time:     Mon, 11 May 2020 14:31:19 +0100
Labels:         aadpodidbinding=ingress-azure
                app=ingress-azure
                pod-template-hash=7c85c7cfb6
                release=ingress-azure
Annotations:    prometheus.io/port: 8123
[...]

Issue #779, opened 2020-03-28
runtime error: invalid memory address or nil pointer dereference : status CrashLoopbackOff


billandr:
**Describe the bug**
I get the following error constantly; not sure exactly when; feels random but I see it more often when I deploy a new image via kubectl set image; but not sure it's related.

What's odd is that I think everything is working fine; the proper routes and rules do get added to the AGW. (but I notice it takes about 3 minutes - after pods have been updated - for the new PODs ip addresses to show up in the backend pool)  But even after they finally get added; I still have that CrashLoopBackOff error.

I've also had everything running and was seeing that error; then have scaled to 0 and then back to 1 in order to restart the POD; and it would take a little while and never get to ready; then finally show the error.  So I suspect something is going wrong when it's syncing with the AGW.  Does the error give any hints? (e.g. [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x12b6ed7] )

This seems related to #629 but not sure.

`ERROR: logging before flag.Parse: I0328 08:25:19.829930       1 main.go:306] Using verbosity level 3 from environment variable APPGW_VERBOSITY_LEVEL
I0328 08:25:19.858639       1 environment.go:168] KUBERNETES_WATCHNAMESPACE is not set. Watching all available namespaces.
I0328 08:25:19.858745       1 main.go:149] App Gateway Details: Subscription: 334721bb-4843-4295-97b5-ea17504c130b, Resource Group: tdp-k8s, Name: agw
I0328 08:25:19.858768       1 auth.go:90] Creating authorizer from Azure Managed Service Identity
I0328 08:25:19.858809       1 httpserver.go:57] Starting API Server on :8123
I0328 08:25:20.224296       1 main.go:191] Ingress Controller will observe all namespaces.
I0328 08:25:20.260472       1 context.go:129] k8s context run started
I0328 08:25:20.260516       1 context.go:168] Waiting for initial cache sync
I0328 08:25:20.360729       1 context.go:176] Initial cache sync done
I0328 08:25:20.360753       1 context.go:177] k8s context run finished
I0328 08:25:20.360871       1 worker.go:35] Worker started
I0328 08:25:20.538911       1 redirects.go:44] Created redirection configuration sslr-fl-9cc0407a3335a7032907821ad24f777d for ([staging.dictionaryproject.org    ],443); not yet linked to a routing rule
[...]

Issue #752, opened 2020-03-04
External-DNS Causes Ingress Controller to Crash (GovCloud AKS)


flyer299:
**Describe the bug**
Note, we are hosted in the VA Gov Cloud.  Not sure if this issue is related or not, just wanted to make sure those who can help are aware.

We have installed the AKS Cluster using the Greenfield Deploy and successfully setup the AGIC using help from GitHub Issues to get a version that can support the Gov Cloud. 

Then we moved on to follow this: https://github.com/kubernetes-sigs/external-dns and more specifically: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure.md

The ingress-azure pod continues to crash and no DNS entries are updated. 

**To Reproduce**
In the VA Gov Cloud:
1) Green Field Deploy
2)  https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure.md

**Ingress Controller details**
* Output of `kubectl describe pod <ingress controller`> . The <ingress controller> pod name can be obtained by running `helm list`. 

```
Name:           ingress-azure-1582659882-84b884bb97-6zfd4
Namespace:      default
Priority:       0
Node:           aks-agentpool-35064155-0/10.0.0.4
Start Time:     Mon, 02 Mar 2020 10:30:52 -0600
Labels:         aadpodidbinding=ingress-azure-1582659882
                app=ingress-azure
                pod-template-hash=84b884bb97
                release=ingress-azure-1582659882
Annotations:    prometheus.io/port: 8123
                prometheus.io/scrape: true
Status:         Running
IP:             10.0.0.11
IPs:            <none>
Controlled By:  ReplicaSet/ingress-azure-1582659882-84b884bb97
Containers:
  ingress-azure:
    Container ID:   docker://7be7ebe06c48beb1b2db480f6a90dc9e5a339b859b75e308020bd817fb910b5a
    Image:          mcr.microsoft.com/azure-application-gateway/kubernetes-ingress:1.0.1-rc2
    Image ID:       docker-pullable://mcr.microsoft.com/azure-application-gateway/kubernetes-ingress@sha256:1543bf5f9b60534d7702ef199dd573ff2d301f3f3f8df96d99da9db7f2f7965a
    Port:           <none>
    Host Port:      <none>
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       Error
      Exit Code:    2
[...]

PR #853, opened 2020-05-14
httpsetting to have https protocol in case port is at 443

3quanfeng:
This PR addresses two issues:
1) When a service uses port 443, this change will initialise the protocol to HTTPS by default. This behaviour can changed by using the backend-protocol annotation.
2) Fix for nil exception when looking for a service.

Fixes: #851, #779, #752, #635, #629

### DIFF
--- a/pkg/appgw/health_probes.go
+++ b/pkg/appgw/health_probes.go
@@ -107,9 +107,16 @@ func (c *appGwConfigBuilder) generateHealthProbe(backendID backendIdentifier) *n
 		probe.Path = to.StringPtr(backendID.Path.Path)
 	}
 
-	// backend protocol "https" now supports to use certificate signed by well-known root certificate as well as trusted self-signed certificate
-	if protocol, _ := annotations.BackendProtocol(backendID.Ingress); protocol == annotations.HTTPS {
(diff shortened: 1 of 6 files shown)


pkg/appgw/health_probes.go
```diff
@@ -107,11 +107,18 @@ func (c *appGwConfigBuilder) generateHealthProbe(backendID backendIdentifier) *n
 		probe.Path = to.StringPtr(backendID.Path.Path)
 	}
 
-	// backend protocol "https" now supports to use certificate signed by well-known root certificate as well as trusted self-signed certificate
-	if protocol, _ := annotations.BackendProtocol(backendID.Ingress); protocol == annotations.HTTPS {
+	if backendID.Backend.ServicePort.IntVal == 443 {
```
akshaysngupta, line 110: can you we use `backendID.Backend.ServicePort.String()` ?
Since this is a union type, it can get int or string as input. Will IntVal work when using string: "443"

3quanfeng: > can you we use `backendID.Backend.ServicePort.String()` ?
> Since this is a union type, it can get int or string as input. Will IntVal work when using string: "443"

Updated.

PR #862, opened 2020-05-18
e2e: add a test for e2e ssl

akshaysngupta:
This PR adds a test e2e ssl using the trusted root certificate annotation. We also test the automatic selection of https when using 443 for both backend http setting and health probe.

I have updated the e2e app gateway with the root certificate by using the az cli. This is not done in the e2e code.


akshaysngupta:
Here is a successful run:
https://dev.azure.com/azure/application-gateway-kubernetes-ingress/_releaseProgress?_a=release-environment-extension&releaseId=265&environmentId=543&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab

PR #867, opened 2020-05-19
Added check to prevent UnhealthyThreshold to go beyond 20

veldman:
When deploying Prometheus Operator on our cluster we run into the problem that the UnhealthyThreshold goes beyond 20 and the ingress controller gives an error: ApplicationGatewayProbeUnhealthyThresholdIsInvalid.

See: 

> error: network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="ApplicationGatewayProbeUnhealthyThresholdIsInvalid" Message="Unhealthy threshold specified for Probe /subscriptions/.../resourceGroups/rg-gateway/providers/Microsoft.Network/applicationGateways/gateway/probes/pb-monitoring-mdf-prometheus-operator-prometheu-cc513868657fb3a5420b340c3bd9625c is not valid. Supported range is [0, 20]." Details=[]'

Currently the code just grabs the `FailureThreshold` from the Probe, but in the case of Prometheus we have no control over it. See: https://github.com/coreos/prometheus-operator/blob/6b1a5276fe42efd355b001cd891cf169b9bd6775/pkg/prometheus/statefulset.go#L651

### DIFF
--- a/pkg/appgw/health_probes.go
+++ b/pkg/appgw/health_probes.go
@@ -148,6 +148,8 @@ func (c *appGwConfigBuilder) generateHealthProbe(backendID backendIdentifier) *n
 			probe.Timeout = to.Int32Ptr(k8sProbeForServiceContainer.TimeoutSeconds)
 		}
 		if k8sProbeForServiceContainer.FailureThreshold != 0 {
+			// UnhealthyThreshold must be in range of 0 - 20, otherwise Application Gateway will not
+			// accept the configuration and all new pods will not be configured.
 			if k8sProbeForServiceContainer.FailureThreshold > 20 {
 				probe.UnhealthyThreshold = to.Int32Ptr(20)
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/appgw/health_probes.go
```diff
@@ -148,7 +148,11 @@ func (c *appGwConfigBuilder) generateHealthProbe(backendID backendIdentifier) *n
 			probe.Timeout = to.Int32Ptr(k8sProbeForServiceContainer.TimeoutSeconds)
 		}
 		if k8sProbeForServiceContainer.FailureThreshold != 0 {
-			probe.UnhealthyThreshold = to.Int32Ptr(k8sProbeForServiceContainer.FailureThreshold)
+			if k8sProbeForServiceContainer.FailureThreshold > 20 {
```
3quanfeng, line 151: could you help add a comment why is 20 specifically?

veldman: Comment added. If you agree with the statement please resolve the converstation.

3quanfeng: Thanks!

PR #870, opened 2020-05-21
e2e: add test case to test backend availability with health probing

3quanfeng:
## Description
- backend will be removed when health probe is unhealthy
-  appgw health probe is determined by the readiness probe defined in the container

### DIFF
--- a/scripts/e2e/cmd/runner/testdata/one-namespace-one-ingress/container-readiness-probe/app.yaml
+++ b/scripts/e2e/cmd/runner/testdata/one-namespace-one-ingress/container-readiness-probe/app.yaml
@@ -1,0 +1,138 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: websocket-deployment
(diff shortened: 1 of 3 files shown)


scripts/e2e/cmd/runner/testdata/one-namespace-one-ingress/container-readiness-probe/app.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - path: /good
+          backend:
+            serviceName: websocket-service
+            servicePort: 80
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: websocket-deployment
```
akshaysngupta, line 74: let's call this unhealthy backend ?

3quanfeng: > let's call this unhealthy backend ?

updated

PR #876, opened 2020-05-21
support install crd by helm3

3quanfeng:
## Description
Now support to install crd by helm 2 as well as helm 3
hook: "crd-install will be unknown and skipped when helm3 is used

## Fixes
crd cannot be installed neither by helm 2 nor helm 3


Wema:
Any idea when this will be available?

PR #882, opened 2020-05-26
e2e: test brownfield setup

3quanfeng:
## Description
- Add e2e test case to test brownfield scenario
- brownfield set up will install non-shared environment first and then reinstalled with shared enabled, non-blacklisted rule is expected to be wiped out, while blacklisted rules shall be retained


scripts/e2e/run-e2e.sh
```diff
@@ -26,22 +26,66 @@ function InstallAGIC() {
     helm repo add staging https://appgwingress.blob.core.windows.net/ingress-azure-helm-package-staging/
     helm repo update
 
-    # AAD pod identity is taking time to assign identity. Timeout is set to 60 sec
+    # delete agic if exist
+    helm delete agic-${version} -n agic || true
```
akshaysngupta, line 30: do we need to delete ?

scripts/e2e/run-e2e.sh
```diff
@@ -26,22 +26,66 @@ function InstallAGIC() {
...
+    kubectl apply -f test-prohibit-backend.yaml
+
+}
+
+function SetupSharedBackend() {
+    
+    # delete agic with share disabled
+    helm delete agic-${version} -n agic || true
+
+    helm repo update
```
akshaysngupta, line 57: no need to upgrade ?

scripts/e2e/test-prohibit-backend.yaml
```diff
@@ -0,0 +1,61 @@
...
+kind: Ingress
+metadata:
+  name: brownfield-blacklist-ingress
+  namespace: test-brownfield-ns
+  annotations:
+    kubernetes.io/ingress.class: azure/application-gateway
+    appgw.ingress.kubernetes.io/backend-path-prefix: "/"
+spec:
+  rules:
+    - host: brownfield-blacklist-ns.host
```
akshaysngupta, line 55: host here doesn't match the hostname in the above .yaml

brownfield-blacklist.host != brownfield-blacklist-ns.host

scripts/e2e/run-e2e.sh
```diff
@@ -26,22 +26,66 @@ function InstallAGIC() {
...
+    --set armAuth.identityResourceID=${identityResourceId} \
+    --set armAuth.identityClientID=${identityClientId} \
+    --set rbac.enabled=true \
+    --timeout 120s \
+    --wait \
+    -n agic \
+    --version ${version}
+
+    # create namespace for testing shared backend
+    kubectl create namespace test-brownfield-ns || true
```
akshaysngupta, line 45: create the namespace as part of the yaml ?

scripts/e2e/run-e2e.sh
```diff
@@ -26,22 +26,66 @@ function InstallAGIC() {
...
+
+    # apply backends to test both non-prohibited and prohibited target
+    kubectl apply -f test-prohibit-backend.yaml
+
+}
+
+function SetupSharedBackend() {
+    
+    # delete agic with share disabled
+    helm delete agic-${version} -n agic || true
```
akshaysngupta, line 55: do we need to delete ? Can you just update the current release ?

scripts/e2e/run-e2e.sh
3quanfeng: removed

scripts/e2e/run-e2e.sh
3quanfeng: updated

scripts/e2e/run-e2e.sh
3quanfeng: removed

PR #755, opened 2020-03-05
retry: keep retrying on AppGW until access is granted

akshaysngupta:
Following is an example where AGIC waits to get access:
```
                                                                                                 Cleanup: delete /home/vsonline/go/src/github.com/Azure/application-gateway-kubernetes-ingress/bin
Compiling...
Build SUCCEEDED
ERROR: logging before flag.Parse: I0311 17:26:12.107653   11457 utils.go:102] Using verbosity level 1 from CLI flag verbosity
Version: 1.0.1-rc3; Commit: 415e854f; Date: 2020-03-11-17:26T+0000
ERROR: logging before flag.Parse: I0311 17:26:12.115781   11466 utils.go:102] Using verbosity level 9 from CLI flag verbosity
ERROR: logging before flag.Parse: I0311 17:26:12.115867   11466 main.go:77] Unable to load cloud provider config:
E0311 17:26:12.118789   11466 context.go:185] Error fetching AGIC Pod (This may happen if AGIC is running in a test environment). Error: resource name may not be empty
I0311 17:26:12.119326   11466 main.go:124] App Gateway Details: Subscription: xxxx, Resource Group: MC_resgp_cluster_westeurope, Name: applicationgateway
I0311 17:26:12.119356   11466 auth.go:37] Creating authorizer from file referenced by environment variable: /home/vsonline/.azure/auth.json
I0311 17:26:12.119378   11466 httpserver.go:57] Starting API Server on :8123
E0311 17:26:12.570168   11466 client.go:184] Unexpected status code '403' while performing a GET on Application Gateway. Please check if AGIC Identity has 'Contributor' access on Application Gateway 'applicationgateway' and 'Reader' access on Application Gateway's Resource Group 'MC_resgp_cluster_westeurope'. Err: network.ApplicationGatewaysClient#Get: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'yyyy' with object id 'yyyy' does not have authorization to perform action 'Microsoft.Network/applicationGateways/read' over scope '/subscriptions/xxxx/resourceGroups/MC_resgp_cluster_westeurope/providers/Microsoft.Network/applicationGateways/applicationgateway' or the scope is invalid. If access was recently granted, please refresh your credentials."
I0311 17:26:12.570216   11466 retry.go:32] Retrying in 10s
[...]

PR #891, opened 2020-05-29
retry: keep retrying on AppGW until access is granted 

akshaysngupta:
## Checklist

## Description

This PR makes AGIC wait forever until it gets access to Application Gateway.
1. added a WaitForGetAccessOnGateway
1. updated configmap to take clientId provided in the aad pod identity. This information will be used to provide full az cli command to use if permission is missing.
1. changed retry function to take -1 as forever.
1. added e2e using role assignment client.

Closed: https://github.com/Azure/application-gateway-kubernetes-ingress/pull/755
I have closed this PR after realising that role assignments are created using objectIDs and AGIC only has clientId which cannot be used to determine if AGIC has contributor access to App Gateway or not. When we try to list the role assignments, we get the whole list of assignments and we cannot filter this list without objectId. Asking users to provide objectId would help but that would increase the variables we are asking the user to provide and making this mandatory might make it harder to upgrade.


akshaysngupta:
E2E passing:
https://dev.azure.com/azure/application-gateway-kubernetes-ingress/_releaseProgress?releaseId=290&_a=release-environment-extension&environmentId=576&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab

pkg/azure/client.go
```diff
@@ -118,50 +114,76 @@ func (az *azClient) SetAuthorizer(authorizer autorest.Authorizer) {
...
+				if response.Response.StatusCode == 403 {
+					e.Code = controllererrors.ErrorApplicationGatewayForbidden
+
+					clientID := "<agic-client-id>"
+					if az.clientID != "" {
+						clientID = az.clientID
+					}
+
+					e.Message += fmt.Sprintf(
+						" You can use 'az role assignment create --role Contributor --scope /subscriptions/%s/resourceGroups/%s --assignee %s' to assign permissions.",
```
3quanfeng, line 152: Should the scope be App-Gateway-ID instead of resource group? resource group only needs reader role

pkg/environment/environment.go
```diff
@@ -20,6 +20,9 @@ const (
 	// CloudProviderConfigLocationVarName is an environment variable name. This file is available on azure cluster.
 	CloudProviderConfigLocationVarName = "AZURE_CLOUD_PROVIDER_LOCATION"
 
+	// ClientIDVarName is an environment variable which stores the client id provided through user assigned identity
+	ClientIDVarName = "AZURE_CLIENT_ID"
```
3quanfeng, line 24: AZURE_UMI_CLIENT_ID?

cmd/appgw-ingress/main.go
```diff
@@ -138,24 +138,27 @@ func main() {
 		azClient.SetAuthorizer(authorizer)
 	}
 
-	if _, err = azClient.GetGateway(); err != nil {
-		if controllererrors.IsErrorCode(err, controllererrors.ErrorApplicationGatewayNotFound) &&
-			env.EnableDeployAppGateway {
+	// check if application gateway exists/have get access
+	// When an AAD SP doesn't have access to the app gateway resource group, then AGIC can't read it's role assignments to look for the needed permission.
```
3quanfeng, line 142: nit: When an AAD SP doesn't have read access to the app gateway resource group...

pkg/environment/environment.go
akshaysngupta: This is env variable defined the by Azure Go SDK. So we have to use this string.
https://github.com/Azure/azure-sdk-for-go#more-authentication-details

pkg/environment/environment.go
```diff
@@ -109,7 +113,7 @@ type EnvVariables struct {
 	AppGwResourceID             string
 	AppGwSubnetID               string
 	AuthLocation                string
-	IngressClass               string
+	IngressClass                string
```
3quanfeng: unrelated change?

pkg/azure/client.go
akshaysngupta: 1) Either we can add this:
```
az role assignment create --role Reader --scope /subscriptions/%s/resourceGroups/%s --assignee %s
az role assignment create --role Contributor --scope <appGatewayId> --assignee %s
```

2) Or we can aim for Contributor to the resource group
a) One command to give access
b) Might be helpful when vnet is in the same resource group. that way, we can use this access to apply route table to the subnet through AGIC without asking for addtitional permission to the Virtual network.

3quanfeng: I think we should go with 1st option, the role assignment should be granted as least as possible unless the contributor role becomes mandatory

scripts/e2e/cmd/runner/environment.go
```diff
@@ -0,0 +1,79 @@
...
+		SubscriptionID:     os.Getenv(SubscriptionIDVarName),
+		ResourceGroupName:  os.Getenv(ResourceGroupNameVarName),
+		AppGwName:          os.Getenv(AppGwNameVarName),
+		KubeConfigFilePath: GetEnvironmentVariable(KubeConfigVarName, "~/.kube/config", nil),
+		ObjectID:           os.Getenv(ObjectIDVarName),
+	}
+}
+
+// GetGroupResourceID returns group's resource id
+func (env *EnvVariables) GetGroupResourceID() string {
```
3quanfeng, line 54: GetResourceGroupID()?

scripts/e2e/cmd/runner/mfu_one_namespace_one_ingress_test.go
```diff
@@ -129,6 +129,66 @@ var _ = Describe("MFU", func() {
...
+
+			// delete the AGIC pod. This will create the pod
+			klog.Info("Deleting AGIC pod")
+			err = deleteAGICPod(clientset)
+			Expect(err).To(BeNil())
+
+			// wait for 30 seconds
+			klog.Info("Wait for 60 seconds")
+			time.Sleep(60 * time.Second)
+
```
3quanfeng, line 159: how can we check if AGIC pod is in ready status?

akshaysngupta: let me address this in a separate PR

3quanfeng: Sounds good, please help track with a task

PR #917, opened 2020-06-17
Configure AzureIngressProhibitedTarget through Helm

### LINKED ISSUE #815: Helm install - decouple appgw.shared Deployment/CRD config from deploying default AzureIngressProhibitedTarget
During IaaC deployment with shared App Gw 3 things happen:
- CRD AzureIngressProhibitedTarget is deployed
- ENV flag is set on Deployment
- Default deny all AzureIngressProhibitedTarget is configured

This makes declarative deployment of cluster scale more complex as deleting objects is imperative and need to modify Helm charts is not convenient.

Would it be possible to set Chart in a way that this default deny all rule is not deployed so I can setup proper exceptions with different decalarative deployment without need to delete something imperatively?

appgw.share flag would deploy CRD definition and ENV flag in Deployment
appgw.share-denyall whould be true by default and would deploy deny-all AzureIngressProhibitedTarget when appgw.share is set 

With this administrator can disable creation of deny-all rule and use declarative way to setup cluster state.

What do you think? Make sense?

Thanks, 

Tomas

askew:
## Checklist

## Description

When deploying AGIC for use in a shared environment ([Multi-cluster / Shared App Gateway](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/setup/install-existing.md#multi-cluster--shared-app-gateway)) you have to deploy separate Kubernetes manifests for the additional `AzureIngressProhibitedTarget` resources, plus you have to manually remove the default `prohibit-all-targets` resource.

This change will allow the prohibited targets configuration via Helm.

In the values file you can add 
``` Yaml
appgw:
  shared: true
  prohibitedTargets:
  - name: protect-myotherapp
    paths:
    - /myotherapp/*
  - name: protect-contoso-prod
    hostname: prod.contoso.com
```

This also makes the `appgw.shared` setting redundant. A further change could be to just check for existence of `appgw.prohibitedTargets`.

## Fixes

#815


akshaysngupta:
Thanks for submitting this PR. It looks good as such.

Will discuss with the team and provide an update.

helm/ingress-azure/templates/crds.yaml
```diff
@@ -4,12 +4,26 @@
...
 kind: AzureIngressProhibitedTarget
 metadata:
-  name: prohibit-all-targets
+  name: {{ .name }}
+  {{- if $watchNamespace }}
+  namespace: {{ $watchNamespace | splitList "," | first }}
+  {{- end }}
 spec:
+  {{- if .hostname }}
+  hostname: {{ .hostname }}
```
akshaysngupta, line 18: I think it would be nice to install "/*" when user doesn't specify custom rules. What do you think ?

askew: I've added the existing `prohibit-all-targets` CRD if `appgw.prohibitedTargets` is missing or empty. This has the existing behaviour if you just specify `appgw.shared=true`

akshaysngupta:
Thanks a lot @askwe!
LGTM

mohanrao:
I tried setting the property --set appgw.prohibitedTargets[0].name=test.com  --set appgw.prohibitedTargets[0].host=test.com but the template didn't looked at it .

LoicAG:
@askew Thanks a lot for this! I was looking for this exact feature, instead of manually deleted the "deny all" CRD :)

@mohanrao you need to set `appgw.prohibitedTargets[0].hostname`, not `appgw.prohibitedTargets[0].host`

PR #958, opened 2020-07-28
Added binding for APPGW_CONFIG_NAME_PREFIX in helm chart

### LINKED ISSUE #957: Use unique identifiers in agw internal resources
I use agic in shared mode to manage and agw for 2 environments deployed using helm. This environments are similar between them because one of the is dev environment and the other is uat environment. The problem comes from the way of the backendpool's names creation. With 2 or more environments those manage similar the workflows (dev,pre,uat,pro...), the backends and http configurations are rewritten all time because the name is the same for every environments.  
In my case, we work on dev environment, and when we have a feature, we promote the environment to uat, and when we have the acceptance of the customer, we promote the environment to production.

Request a name during the setup of agic and append that name to the agw items name. This change allows to manage the same agw from several environment clusters and if you don't provide any name, agic works like now.

I saw that the environment variable APPGW_CONFIG_NAME_PREFIX is used to set the prefix in all names but is not configurable with helm. Maybe be great idea allow configure it in the chart deployment

JorTurFer:
## Checklist

## Description
When work with agic in shared mode, is a normal scenario when you have 2 or more environment behind the same agw and promote from one to the next (dev -> uat -> prod).  In this case, the namespaces and services have the same naming, so all the agw internal resources are in conflict between them in the different environments. 
Agic supports the use of prefix that is added in each resource but that setting is not editable using helm. 
This PR added support to manage the prefix in order to allow share the same agw between different clusters those have the same application in different environments.

## Fixes
#957

### DIFF
--- a/docs/helm-values-documenation.md
+++ b/docs/helm-values-documenation.md
@@ -12,7 +12,7 @@
 | `appgw.name` | | Name of the Application Gateway. Example: `applicationgatewayd0f0` |
 | `appgw.environment`| `AZUREPUBLICCLOUD` | Specify which cloud environment. Possbile values: `AZURECHINACLOUD`, `AZUREGERMANCLOUD`, `AZUREPUBLICCLOUD`, `AZUREUSGOVERNMENTCLOUD` |
 | `appgw.shared` | false | This boolean flag should be defaulted to `false`. Set to `true` should you need a [Shared App Gateway](setup/install-existing.md#multi-cluster--shared-app-gateway). |
-| `appgw.prefix` | No prefix if empty | Prefix that should be used in the naming of the Application Gateway's sub-resources|
+| `appgw.subResourceNamePrefix` | No prefix if empty | Prefix that should be used in the naming of the Application Gateway's sub-resources|
 | `kubernetes.watchNamespace` | Watches all if empty | Specify the name space, which AGIC should watch. This could be a single string value, or a comma-separated list of namespaces. |
 | `kubernetes.rbac` | false | Specify true if kubernetes cluster is rbac enabled |
 | `armAuth.type` | | could be `aadPodIdentity` or `servicePrincipal` |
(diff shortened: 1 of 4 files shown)


docs/helm-values-documenation.md
```diff
@@ -12,6 +12,7 @@
 | `appgw.name` | | Name of the Application Gateway. Example: `applicationgatewayd0f0` |
 | `appgw.environment`| `AZUREPUBLICCLOUD` | Specify which cloud environment. Possbile values: `AZURECHINACLOUD`, `AZUREGERMANCLOUD`, `AZUREPUBLICCLOUD`, `AZUREUSGOVERNMENTCLOUD` |
 | `appgw.shared` | false | This boolean flag should be defaulted to `false`. Set to `true` should you need a [Shared App Gateway](setup/install-existing.md#multi-cluster--shared-app-gateway). |
+| `appgw.prefix` | No prefix if empty | Prefix that should be used in the naming of the agw internal resources|
```
akshaysngupta, line 15: agw internal resources -> Application Gateway's sub-resources

JorTurFer:
> Added a minor comment
Done! :)

docs/helm-values-documenation.md
```diff
@@ -12,6 +12,7 @@
 | `appgw.name` | | Name of the Application Gateway. Example: `applicationgatewayd0f0` |
 | `appgw.environment`| `AZUREPUBLICCLOUD` | Specify which cloud environment. Possbile values: `AZURECHINACLOUD`, `AZUREGERMANCLOUD`, `AZUREPUBLICCLOUD`, `AZUREUSGOVERNMENTCLOUD` |
 | `appgw.shared` | false | This boolean flag should be defaulted to `false`. Set to `true` should you need a [Shared App Gateway](setup/install-existing.md#multi-cluster--shared-app-gateway). |
+| `appgw.prefix` | No prefix if empty | Prefix that should be used in the naming of the Application Gateway's sub-resources|
```
akshaysngupta, line 15: Hey sorry for late comment, do you think `subResourceNamePrefix` would be a more clear name ?

JorTurFer: Maybe will be better name, let me a moment to change it :)

JorTurFer: Hi @akshaysngupta 
Done! :)

PR #963, opened 2020-08-02
Added the use of the sub resources prefix also in the ssl certificate name

### LINKED ISSUE #957: Use unique identifiers in agw internal resources

JorTurFer:
## Checklist
- [X] The title of the PR is clear and informative
- [X] If applicable, the changes made in the PR have proper test coverage
- [X] Issues addressed by the PR are mentioned in the description followed by `Fixes`.

## Description
After some tries, I found that #958 is not enough because the ssl certificates in agw is created under the name "namespace-secretname". Using the prefix all subresources use it but the certificates still be in conflict because them don't use the prefix and one environment overrides the certificate from other environments.
This PR added the use of the agPrefix in the name of the certificate in order to allow certificates with same namespace and secret for every different environment.
Also added "cert-" as constant in certificate naming in order to be consistent with other naming rules

## Fixes
#957


pkg/appgw/internaltypes.go
```diff
@@ -116,9 +117,9 @@ func (s secretIdentifier) secretKey() string {
 func (s secretIdentifier) secretFullName() string {
 	// in case of referring ssl certificate from agic annotation
 	if len(s.Namespace) == 0 {
-		return s.Name
+		return fmt.Sprintf("%v%v-%v", agPrefix, prefixSslCertificate, s.Name)
```
3quanfeng, line 120: One use case about null namespace here, i.e. [appgw-ssl-certificate](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/annotations.md#appgw-ssl-certificate), is that the secret will be pre-installed on AppGw with a user-specified name. AppGw will not be able to reference the right sslCertificate resource if the prefix added here.

JorTurFer: Hi @3quanfeng ,
You are right! I understood bad the way as agic works but your commentary about it is very logic. I will change it during the day

JorTurFer: Done @3quanfeng ! :)

PR #975, opened 2020-08-10
fix: duplicate URL Path Map names w/ certain Ingress YAML

akshaysngupta:
## Checklist

## Description

This PR fixes a bug in the naming of the path rules in the URL path maps in the Application Gateway config. This was reported as bug in an CRI.
Currently, when an ingress contains multiple ingress rules but with same `host`, the paths are merged into the same URL path map but with the same name which causes the validation to fail on the gateway.

**Example**
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: aspnetapp
  annotations:
    appgw.ingress.kubernetes.io/backend-path-prefix: "/"
spec:
  rules:
  - http:
      paths:
      - path: "/picking/*"
        backend:
          serviceName: aspnetapp
          servicePort: 80
  - http:
       paths:
       - path: "/loading/*"
         backend:
           serviceName: aspnetapp2
           servicePort: 80
```
**Error**
`error: network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequestFormat" Message="Cannot parse the request." Details=[{"code":"DuplicateResourceName","message":"Resource /subscriptions//resourceGroups//providers/Microsoft.Network/applicationGateways//urlPathMaps/ has two child resources with the same name (pr-fulfilment-point-services-fps-ingress-0)."},{"code":"DuplicateResourceName","message":"Resource /subscriptions//resourceGroups//providers/Microsoft.Network/applicationGateways//urlPathMaps/ has two child resources with the same name (pr-fulfilment-point-services-fps-ingress-0)."}]`

**Note**: Will add E2E in a subsequent PR.

### DIFF
--- a/functional_tests/functional_test.go
+++ b/functional_tests/functional_test.go
@@ -167,7 +167,7 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
 				annotations.IngressClassKey: annotations.ApplicationGatewayIngressClass,
 			},
 			Namespace: tests.Namespace,
-			Name:      tests.Name,
+			Name:      tests.Name + "FooBazNoTLS",
 		},
 	}
 
@@ -200,7 +200,7 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
 				annotations.HostNameExtensionKey: "foo.baz",
 			},
 			Namespace: tests.Namespace,
-			Name:      tests.Name,
+			Name:      tests.Name + "FooBazNoTLSHostNameFromAnnotation",
 		},
 	}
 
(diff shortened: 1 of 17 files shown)


pkg/appgw/requestroutingrules_test.go
```diff
@@ -159,12 +159,12 @@ var _ = Describe("Test routing rules generations", func() {
 			Expect(len(*generatedPathMap.PathRules)).To(Equal(2))
```
3quanfeng, line 159: One more check to identify the uniqueness of the pathRuleName?

functional_tests/functional_test.go
```diff
@@ -167,7 +167,7 @@ var _ = ginkgo.Describe("Tests `appgw.ConfigBuilder`", func() {
 				annotations.IngressClassKey: annotations.ApplicationGatewayIngressClass,
 			},
 			Namespace: tests.Namespace,
-			Name:      tests.Name,
+			Name:      tests.Name + "FooBazNoTLS",
```
3quanfeng, line 170: Is there any specific reason to add suffix here?

akshaysngupta: yes, some test use multiple ingresses and it would be not be right to have the same names for them.

PR #990, opened 2020-09-01
fix: choose backend port in case backend port name is resolved to multiple backend ports

abjai:
Updating logic to choose single backend port in case backend port name is resolved to multiple backend ports retrieved by endpoints

Currently, if service target endpoint is not provided, endpoints api is called to see what are the backend ports and in case multiple backend ports exists with same name, AGIC stops processing the update and PUT will not be issued to AppGw. Only way to fix this problem is deleting the ingress which lead to this problem.

After this changes, in case there are multiple target backend ports, smallest backend port is chosen to create backend http settings.

## Checklist

## Description

Currently, if service target endpoint is not provided, endpoints api is called to see what are the backend ports and in case multiple backend ports exists with same name, AGIC stops processing the update and PUT will not be issued to AppGw. Only way to fix this problem is deleting the ingress which lead to this problem.

## Fixes

After this changes, in case there are multiple target backend ports, smallest backend port is chosen to create backend http settings.

### DIFF
--- a/pkg/appgw/backendhttpsettings.go
+++ b/pkg/appgw/backendhttpsettings.go
@@ -167,10 +167,6 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 
 	// enforce single pair relationship between service port and backend port
 	for backendID, serviceBackendPairs := range serviceBackendPairsMap {
-		if len(serviceBackendPairs) > 1 {
-			// more than one possible backend port exposed through ingress
-			glog.V(5).Infof("service:port [%s:%s] has more than one service-backend port binding", backendID.serviceKey(), backendID.Backend.ServicePort.String())
-		}
 
 		// in case there are multiple backend ports found, using the smallest port in http setting
 		var backendport Port
@@ -181,6 +177,19 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 				uniquePair = k
 				backendport = k.BackendPort
 			}
+		}
+
+		if len(serviceBackendPairs) > 1 {
+
+			// more than one possible backend port exposed through ingress
+			e := controllererrors.NewErrorf(
+				controllererrors.ErrorMultipleServiceBackendPortBinding,
+				"service:port [%s:%s] has more than one service-backend port binding which is not an ideal scenario, choosing the smallest service-backend port %d",
+				backendID.serviceKey(), backendID.Backend.ServicePort.String(), backendport,
+			)
+
+			c.recorder.Event(backendID.Ingress, v1.EventTypeWarning, events.ReasonPortResolutionError, e.Error())
+			glog.Errorf(e.Error())
 		}
 
 		finalServiceBackendPairMap[backendID] = uniquePair
@@ -194,8 +203,6 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 	for _, backend := range httpSettingsCollection {
 		httpSettings = append(httpSettings, backend)
 	}
-
-	glog.V(5).Infof("httpsettingsCount %d", len(httpSettings))
 
 	c.mem.settings = &httpSettings
 	c.mem.settingsByBackend = &backendHTTPSettingsMap
--- a/pkg/appgw/backendhttpsettings.go
+++ b/pkg/appgw/backendhttpsettings.go
@@ -169,20 +169,18 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 	for backendID, serviceBackendPairs := range serviceBackendPairsMap {
 		if len(serviceBackendPairs) > 1 {
 			// more than one possible backend port exposed through ingress
-			e := controllererrors.NewErrorf(
-				controllererrors.ErrorMultipleServiceBackendPortBinding,
-				"service:port [%s:%s] has more than one service-backend port binding",
-				backendID.serviceKey(), backendID.Backend.ServicePort.String(),
-			)
-			c.recorder.Event(backendID.Ingress, v1.EventTypeWarning, events.ReasonPortResolutionError, e.Error())
-			glog.Warning(e.Error())
-			return nil, nil, nil, e
+			glog.V(5).Infof("service:port [%s:%s] has more than one service-backend port binding", backendID.serviceKey(), backendID.Backend.ServicePort.String())
 		}
 
-		// At this point there will be only one pair
+		// in case there are multiple backend ports found, using the smallest port in http setting
+		var backendport Port
+		backendport = 65536
 		var uniquePair serviceBackendPortPair
 		for k := range serviceBackendPairs {
-			uniquePair = k
+			if k.BackendPort <= backendport {
+				uniquePair = k
+				backendport = k.BackendPort
+			}
 		}
 
 		finalServiceBackendPairMap[backendID] = uniquePair
@@ -196,6 +194,8 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 	for _, backend := range httpSettingsCollection {
 		httpSettings = append(httpSettings, backend)
 	}
+
+	glog.V(5).Infof("httpsettingsCount %d", len(httpSettings))
 
 	c.mem.settings = &httpSettings
 	c.mem.settingsByBackend = &backendHTTPSettingsMap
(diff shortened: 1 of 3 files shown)


pkg/appgw/backendhttpsettings.go
```diff
@@ -197,6 +195,8 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 		httpSettings = append(httpSettings, backend)
 	}
 
+	glog.V(5).Infof("httpsettingsCount %d", len(httpSettings))
```
abjai, line 198: remove

pkg/tests/fixtures.go
```diff
@@ -424,6 +424,68 @@ func NewServiceFixture(servicePorts ...v1.ServicePort) *v1.Service {
 	}
 }
 
+// NewEndpointsFixture makes a new endpoint for testing
```
3quanfeng, line 427: // NewEndpointsFixtureWithSameNameMultiplePorts makes ...

pkg/appgw/backendhttpsettings.go
```diff
@@ -169,20 +169,18 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
...
 			// more than one possible backend port exposed through ingress
-			e := controllererrors.NewErrorf(
-				controllererrors.ErrorMultipleServiceBackendPortBinding,
-				"service:port [%s:%s] has more than one service-backend port binding",
-				backendID.serviceKey(), backendID.Backend.ServicePort.String(),
-			)
-			c.recorder.Event(backendID.Ingress, v1.EventTypeWarning, events.ReasonPortResolutionError, e.Error())
-			glog.Warning(e.Error())
-			return nil, nil, nil, e
+			glog.V(5).Infof("service:port [%s:%s] has more than one service-backend port binding", backendID.serviceKey(), backendID.Backend.ServicePort.String())
```
akshaysngupta, line 172: we should raise this as an error in the logs. Would make sense to let user know.
`glog.Errorf`

pkg/appgw/backendhttpsettings.go
```diff
@@ -169,20 +169,18 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
...
 
-		// At this point there will be only one pair
+		// in case there are multiple backend ports found, using the smallest port in http setting
+		var backendport Port
+		backendport = 65536
 		var uniquePair serviceBackendPortPair
 		for k := range serviceBackendPairs {
-			uniquePair = k
+			if k.BackendPort <= backendport {
+				uniquePair = k
```
akshaysngupta, line 181: should we log here about how we are making the decision ?

akshaysngupta:
@abjai we should also add an E2E test to this PR. Let's discuss that once we get this in.

pkg/appgw/backendhttpsettings_test.go
```diff
@@ -123,3 +125,60 @@ var _ = Describe("Test the creation of Backend http settings from Ingress defini
 
 	})
 })
+
+var _ = Describe("Test the creation of Backend http settings from Ingress definition with target port in Service contains a name which resolved to multiple ports of PODs", func() {
```
akshaysngupta, line 129: I am not sure if a new `Describe` block overrides the previous `Describe` block.
Can you check if the test described above are executing as well ? Try running the test using `-v` to increase the verbosity.
Also, see if you can move this test into the Describe block above.

pkg/appgw/backendhttpsettings.go
akshaysngupta: we should also raise an k8s event on the ingress object on which we found the problem. You can find ample examples in the code to raise an event.

pkg/appgw/backendhttpsettings.go
```diff
@@ -169,20 +169,18 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
 	for backendID, serviceBackendPairs := range serviceBackendPairsMap {
```
3quanfeng, line 169: At line 127 above, targetPortsResolved := c.resolvePortName(sp.Name, &backendID)
we should use sp.TargetPort instead of sp.Name
sp.TargetPort is supposed to be used to compare with endpoint port name

pkg/appgw/backendhttpsettings.go
abjai: We used to raise an event when we encountered such situation, i will be reusing the same code.

pkg/appgw/backendhttpsettings.go
abjai: @3quanfeng i think you are correct. Looks like a bug in our code.
@akshaysngupta  can you confirm?

pkg/appgw/backendhttpsettings_test.go
abjai: I tried to test it in verbose mode and also in debug mode and i could see that both the describe were running and overall test count increased from 152 to 153.

New test scenario needed different endpoint because of which i added a new describe. However i could see that if i change the endpoints in existing describe, there is no impact of the exisiting context. Hence changing the existing describe and moving the test.

abjai: @akshaysngupta @3quanfeng I can see that 3 tests are consistently failing in my local system with/without my changes. Is it expected?

[Fail] Tests `appgw.ConfigBuilder` [It] should have found the pfx certificate 
D:/Aznet/AGIC/application-gateway-kubernetes-ingress/pkg/appgw/public_and_private_ip_test.go:241

[Fail] Tests `appgw.ConfigBuilder` Tests Application Gateway config creation [It] Should have created correct App Gateway config JSON blob 
D:/Aznet/AGIC/application-gateway-kubernetes-ingress/pkg/appgw/public_and_private_ip_test.go:305

[Fail] Tests `appgw.ConfigBuilder` Tests Ingress Controller TLS [It] Should be able to create Application Gateway Configuration from Ingress with TLS.    
D:/Aznet/AGIC/application-gateway-kubernetes-ingress/pkg/appgw/appgw_test.go:652

akshaysngupta: Those test are dependent on openssl. So if you are running on Windows, you need to install openssl manually.

pkg/appgw/backendhttpsettings.go
akshaysngupta: that is what you would think.

But turns out that kubernetes matches the serverPortName with the container port after resolving.

akshaysngupta: We actually had to fix this because of this.
https://github.com/Azure/application-gateway-kubernetes-ingress/pull/393

3quanfeng: If EndpointPort name needs to match the 'name' field in the corresponding ServicePort, can we add such a comment to the code?

pkg/appgw/backendhttpsettings.go
```diff
@@ -167,22 +167,29 @@ func (c *appGwConfigBuilder) getBackendsAndSettingsMap(cbCtx *ConfigBuilderConte
...
+		}
+
 		if len(serviceBackendPairs) > 1 {
+
 			// more than one possible backend port exposed through ingress
 			e := controllererrors.NewErrorf(
 				controllererrors.ErrorMultipleServiceBackendPortBinding,
-				"service:port [%s:%s] has more than one service-backend port binding",
-				backendID.serviceKey(), backendID.Backend.ServicePort.String(),
+				"service:port [%s:%s] has more than one service-backend port binding which is not an ideal scenario, choosing the smallest service-backend port %d",
```
akshaysngupta, line 187: can  we also print the other ports here ?

abjai: You mean endpoint ports matched with the service endpoint name in case of multiple ports?

pkg/tests/fixtures.go
```diff
@@ -424,6 +424,68 @@ func NewServiceFixture(servicePorts ...v1.ServicePort) *v1.Service {
...
+				// because they have not yet finished starting, have recently failed a readiness check,
+				// or have recently failed a liveness check.
+				// +optional
+				NotReadyAddresses: []v1.EndpointAddress{},
+				// Port numbers available on the related IP addresses.
+				// +optional
+				Ports: []v1.EndpointPort{
+					{
+						Protocol: v1.ProtocolTCP,
+						Name:     "--service-https-port--",
```
3quanfeng, line 460: we have the variable ServiceHTTPPort defined already, we can reuse it here as port Name

3quanfeng (approved):

PR #991, opened 2020-09-03
pathrules: merge path rules such that duplicate paths are ignored 

deeksha-microsoft:
## Checklist

## Description
Merge path based rules such that
1.  duplicate paths are ignored
2.  ingress error message is logged with details of the service having duplicate path


pkg/appgw/requestroutingrules_test.go
```diff
@@ -447,4 +447,82 @@ var _ = Describe("Test routing rules generations", func() {
...
+		service1 := tests.NewServiceFixture(*tests.NewServicePortsFixture()...)
+		service2 := tests.NewServiceFixture(*tests.NewServicePortsFixture()...)
+		service2.Name = "testService"
+		endpoint2.Name = service2.Name
+		// 2 path based rules with path - /api1, /api2
+		ingressPathBased := tests.NewIngressFixture()
+		ingressPathBased.Annotations[annotations.SslRedirectKey] = "false"
+		backendBasic := tests.NewIngressBackendFixture(service2.Name, 80)
+		// Adding duplicate path /api for a different service in same ingress
+		ruleApi := tests.NewIngressRuleFixture(tests.Host, "/api1", *backendBasic)
```
3quanfeng, line 464: golint failure: var ruleApi should be ruleAPI

pkg/appgw/requestroutingrules_test.go
```diff
@@ -447,4 +447,82 @@ var _ = Describe("Test routing rules generations", func() {
 			Expect(len(*configBuilder.appGw.URLPathMaps)).To(Equal(0))
 		})
 	})
+
+	FContext("test path-based rule with 1 ingress and multiple service having with same paths", func() {
```
3quanfeng, line 451: Remove Focus

pkg/appgw/requestroutingrules.go
```diff
@@ -365,15 +365,35 @@ func (c *appGwConfigBuilder) mergePathMap(existingPathMap *n.ApplicationGatewayU
...
+			}
+		}
+		if addRuleToMergeList {
+			mergedPathRules = append(mergedPathRules, pathRule)
+		}
 	}
+
+	if len(mergedPathRules) > 0 {
+		existingPathMap.PathRules = &mergedPathRules
+	}
```
deeksha-microsoft, line 396: Added this condition because some tests required `existingPathMap.PathRules` to be nil in case of 0 path rules. Thought on changing test cases vs keeping this condition?

pkg/appgw/requestroutingrules.go
```diff
@@ -365,15 +365,35 @@ func (c *appGwConfigBuilder) mergePathMap(existingPathMap *n.ApplicationGatewayU
...
+		allPathRules = append(*existingPathMap.PathRules, *pathMapToMerge.PathRules...)
+	}
+
+	// we want to ensure that there are only unique paths in the url path map
+	pathMap := make(map[string]n.ApplicationGatewayPathRule)
+	for _, pathRule := range allPathRules {
+		addRuleToMergeList := true
+		for _, path := range *pathRule.Paths {
+			_, exists := pathMap[path]
+			if exists {
```
3quanfeng, line 382: nit: more "go" way
if _, exists := pathMap[path]; exists {
    //do something here
}

pkg/appgw/requestroutingrules.go
```diff
@@ -365,15 +365,35 @@ func (c *appGwConfigBuilder) mergePathMap(existingPathMap *n.ApplicationGatewayU
...
+			} else {
+				pathMap[path] = pathRule
+			}
+		}
+		if addRuleToMergeList {
+			mergedPathRules = append(mergedPathRules, pathRule)
+		}
 	}
+
+	if len(mergedPathRules) > 0 {
```
3quanfeng, line 394: I doubt that this condition will always be satisfied in your code...

deeksha-microsoft: Not when there are absolutely no path rules. Like in test case requestroutingrule_test -> "test waf policy is not configured in rule path",
it is because of this test case that this condition came into play

pkg/appgw/requestroutingrules.go
deeksha-microsoft: @3quanfeng

pkg/appgw/requestroutingrules.go
3quanfeng: > Not when there are absolutely no path rules. Like in test case requestroutingrule_test -> "test waf policy is not configured in rule path",
> it is because of this test case that this condition came into play
I think allPathRules in your code will always contain at least one rule...which makes mergedPathRules include at least one path rule
when is all

deeksha-microsoft: allPathRules is created from input pathMapToMerge.PathRules, and it can be of  length also

pkg/appgw/requestroutingrules.go
```diff
@@ -365,15 +365,35 @@ func (c *appGwConfigBuilder) mergePathMap(existingPathMap *n.ApplicationGatewayU
 		existingPathMap.DefaultBackendAddressPool = nil
 		existingPathMap.DefaultBackendHTTPSettings = nil
 	}
-	if pathMapToMerge.PathRules == nil || len(*pathMapToMerge.PathRules) == 0 {
```
akshaysngupta, line 368: is there a reason this check was removed ?

pkg/appgw/requestroutingrules_test.go
```diff
@@ -463,4 +463,85 @@ var _ = Describe("Test routing rules generations", func() {
...
+		service2 := tests.NewServiceFixture(*tests.NewServicePortsFixture()...)
+		service2.Name = testServiceName
+
+		// 2 path based rules with path - /api1, /api2
+		ingressPathBased := tests.NewIngressFixture()
+		ingressPathBased.Annotations[annotations.SslRedirectKey] = "false"
+		backendBasic := tests.NewIngressBackendFixture(service2.Name, 80)
+
+		// Adding duplicate path /api for a different service in same ingress
+		duplicatePathRule := tests.NewIngressRuleFixture(tests.Host, "/api1", *backendBasic)
```
akshaysngupta, line 483: can you add additional cases to this test covering:
1. duplicate path in same rule under different path [not covered]
1. duplicate path in different rules [covered]
1. duplicate path in different ingresses [not covered]

akshaysngupta (changes requested):
can you please add additional test cases?

pkg/appgw/requestroutingrules_test.go
deeksha-microsoft: Done

pkg/appgw/requestroutingrules.go
deeksha-microsoft: We added a similar condition in a new if block, but yeah makes sense to preserve the original condition.

akshaysngupta (approved):
LGTM! Thanks @deeksha-microsoft 😄

akshaysngupta:
TODO: E2E

PR #1064, opened 2020-11-21
Added support for tolerations and affinities (#1055)

### LINKED ISSUE #1055: Affinities, taints and liveness/readiness configuration with Helm
We would like to configure some deployment attributes via values on Helm:
- Affinities
- Node taints
- Liveness/Readiness timeouts

Currently, this is not possible because values are hardcoded. In the case of Liveness/Readiness timeouts, it could be enough if you increase these values because the defaults are so demanding for AKS. For example, we think values used in this other Azure project (https://github.com/Azure/aad-pod-identity/blob/master/charts/aad-pod-identity/templates/mic-deployment.yaml) are better.

JorTurFer:
## Checklist

## Description
This closes #1055 adding support to set the affinity and the tolerations of the deployment. **This PR also group `nodeSelector` value with `toleration` and `affinity` under `agic` value group.**

## Fixes
#1055


akshaysngupta:
* Would it make sense to group these under existing `kubernetes` key instead of `agic`?
* This is a breaking change for `nodeSelector`. We should support both `nodeSelector` and `kubernetes.nodeSelector`.

JorTurFer:
Yes,
You are right, this values could be better under `kubernetes`. I will prepare the changes now.
I will also add support for both `nodeSelector`

akshaysngupta (approved):
Thank you @JorTurFer !

Issue #893, opened 2020-05-30
Ingress controller not configuring App Gateway


andrewCluey:
**Describe the bug**
Deployed the AKS cluster and APP gateway following 'tutorial' steps here (https://docs.microsoft.com/en-us/azure/developer/terraform/create-k8s-cluster-with-aks-applicationgateway-ingress).

Ingress controller helm seemingly deployed fine, no errors reported at deployment.

running ```kubectl get pods``` shows the following.
```
NAME                                       READY   STATUS    RESTARTS   AGE
ingress-azure-1590841283-989fd77f8-wdtkv   0/1     Running   4          37m
```
Running ```kubectl apply -f aspnetapp.yaml (as described in the tutorial and the troubleshooting guide) seems to go fine, but nothing is actually configured on the app gateway.

App gateway logs show the following:
```
I0530 13:01:10.358915       1 environment.go:210] KUBERNETES_WATCHNAMESPACE is not set. Watching all available namespaces.
I0530 13:01:10.359109       1 main.go:124] App Gateway Details: Subscription: 7df4fea2-d719-4abe-xxxxxxx, Resource Group: aci, Name: AG01
I0530 13:01:10.359127       1 auth.go:46] Creating authorizer from Azure Managed Service Identity
I0530 13:01:10.359252       1 httpserver.go:57] Starting API Server on :8123
```

Rather than the expected:
```
I0927 22:34:51.281437       1 process.go:156] Applied Application Gateway config in 20.461335266s
I0927 22:34:51.281585       1 process.go:165] cache: Updated with latest applied config.
I0927 22:34:51.282342       1 process.go:171] END AppGateway deployment
```

Also, describing the ingress pod gets the following:
```
Name:           ingress-azure-1590843666-c89688c49-drccn
Namespace:      default
Priority:       0
Node:           aks-agentpool-35064155-vmss000001/10.0.0.20
Start Time:     Sat, 30 May 2020 13:01:08 +0000
Labels:         aadpodidbinding=ingress-azure-1590843666
                app=ingress-azure
                pod-template-hash=c89688c49
                release=ingress-azure-1590843666
Annotations:    prometheus.io/port: 8123
                prometheus.io/scrape: true
Status:         Running
IP:             10.0.0.22
IPs:            <none>
Controlled By:  ReplicaSet/ingress-azure-1590843666-c89688c49
Containers:
  ingress-azure:
[...]

Issue #763, opened 2020-03-09
Redirect rule / Cert-manager TLS secret race condition


DenisBiondic:
**Describe the bug**
A sporadic error occurs in which the redirect rule will not be create in app gw. 

Issue is essentialy that the 443 redirect rule is not created because the secret could not be found (but the secret is there, created via cert-manager without any issues): 

...
I0309 07:34:45.683350       1 event.go:258] Event(v1.ObjectReference{Kind:"Ingress", Namespace:"traffic-health-endpoint", Name:"traffic-health-ingress-public-secure", UID:"58a0146b-61d8-11ea-8a77-f62f0564b24c", APIVersion:"extensions/v1beta1", ResourceVersion:"767058", FieldPath:""}): type: 'Warning' reason: 'SecretNotFound' Unable to find the secret associated to secretId: [traffic-health-endpoint/traffic-health-ingress-public-secure-tls]
....

E0309 07:34:45.683503       1 requestroutingrules.go:255] Will not attach default redirect to rule; SSL Redirect does not exist:

DenisBiondic:
Similar to this: https://github.com/Azure/application-gateway-kubernetes-ingress/issues/728 but I am unsure if it's the same issue.

flyer299:
@DenisBiondic , What version of the AGIC are you running?   We are running on kubernetes-ingress:1.0.1-rc2 and are having a very similar if not the exact same issue.

flyer299:
I tried upgrading to 1.0.1-rc3  (not that I expected it result in anything different because the PR doesn't have anything specific for this).   And it didn't fix it.  But now I am running on RC3 for a while.

I am stuck on RC2 or RC3 because I am running in the Gov Cloud.

DenisBiondic:
We are running 

helmchart: "application-gateway-kubernetes-ingress/ingress-azure"
version: "1.0.0"

andres-de-castro:
I was able to work around this by changing the name of the secret in the defined ingress.

thomas-riccardi:
We are running helm chart 1.2.0-rc2 and got the same issue:
- cert-manager with let's encrypt
- first deployment of Ingress
- cert-manager took 6min to issue the certificate (i.e. create the k8s tls secret with it)
- it's as if AGIC stopped trying in loop to update the Application Gateway after 5min

Last logs:
```
I0519 18:11:38.724581       1 mutate_app_gateway.go:177] BEGIN AppGateway deployment
[...]

Issue #728, opened 2020-02-03
Possible race condition between processing Ingress and Secret resource


DenisBiondic:
**Describe the bug**
When using the application gateway with cert-manager, a random issue occurs where the 443 port / route will not be mapped in the application gateway. The application gateway ingress controller simply reports "secret is malformed, tls.key or tls.crt is not defined" (check logs below). Restarting the controller resolves the issue, which seems to me that the watching for these tls secrets is not implemented properly, there is a timing issue somewhere.

Perhaps someone else had a similar issue?

**To Reproduce**
-- no reliable steps found yet, occurs sporadically, every 1 in 10 deployments --

**Ingress Controller details**
Controller Version **1.0.0**

I0129 12:47:24.645533       1 mutate_app_gateway.go:150] cache: Config has NOT changed! No need to connect to ARM.
I0129 13:49:26.780663       1 mutate_app_gateway.go:150] cache: Config has NOT changed! No need to connect to ARM.
I0129 13:49:45.001575       1 mutate_app_gateway.go:150] cache: Config has NOT changed! No need to connect to ARM.
I0129 13:49:52.247375       1 mutate_app_gateway.go:154] BEGIN AppGateway deployment
E0129 13:49:52.376453       1 secretstore.go:75] secret [-] **is malformed, tls.key or tls.crt is not defined**
I0129 13:50:12.849423       1 mutate_app_gateway.go:182] Applied App Gateway config in 20.60201513s
I0129 13:50:12.849454       1 mutate_app_gateway.go:198] cache: Updated with latest applied config.
I0129 13:50:12.850154       1 mutate_app_gateway.go:203] END AppGateway deployment
I0129 13:50:13.957389       1 mutate_app_gateway.go:154] BEGIN AppGateway deployment
I0129 13:50:34.413684       1 mutate_app_gateway.go:182] Applied App Gateway config in 20.456241831s
I0129 13:50:34.413707       1 mutate_app_gateway.go:198] cache: Updated with latest applied config.
I0129 13:50:34.414836       1 mutate_app_gateway.go:203] END AppGateway deployment
I0129 13:50:35.556534       1 mutate_app_gateway.go:150] cache: Config has NOT changed! No need to connect to ARM.
E0131 09:12:05.004845       1 secretstore.go:75] secret [-] **is malformed, tls.key or tls.crt is not defined**
I0131 09:12:05.343003       1 mutate_app_gateway.go:154] BEGIN AppGateway deployment
[...]

PR #1070, opened 2020-11-26
fix: update ingressSecretMap even when secret is malformed

akshaysngupta:
## Checklist

## Description
This PR fixes an issue in processing ingress event with TLS secret.
When ingress is processed, the referenced TLS secret is mapped to ingress and saved in a ingress-secret map.

In specific case where secret is malformed,, this step is skipped due to the `continue` statement. So later, when the certificate is fixed, it doesn't cause an event on AGIC and doesn't update the Application Gateway.
Secret is malformed in case customer is using a certificate manager which creates an empty secret initially and then fills the cert crt and key after generating them.

This is the place where secret handler checks if the secret is in the map or not.
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/6bc8764eb2ecff4a0c6ce96b9124ea36c7e4daa5/pkg/k8scontext/secrets_handlers.go#L55

## Fixes
Closes https://github.com/Azure/application-gateway-kubernetes-ingress/issues/893
Closes https://github.com/Azure/application-gateway-kubernetes-ingress/issues/728
Closes https://github.com/Azure/application-gateway-kubernetes-ingress/issues/763

### DIFF
--- a/pkg/k8scontext/secretstore.go
+++ b/pkg/k8scontext/secretstore.go
@@ -12,7 +12,6 @@ import (
 	"os/exec"
 	"sync"
 
-	"github.com/golang/glog"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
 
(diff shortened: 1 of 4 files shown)


pkg/k8scontext/secretstore.go
```diff
@@ -138,18 +125,15 @@ func (s *SecretsStore) ConvertSecret(secretKey string, secret *v1.Secret) error
 
 	// if openssl exited with an error or the output is empty, report error
 	if err := cmd.Run(); err != nil || len(cout.Bytes()) == 0 {
-		e := controllererrors.NewErrorWithInnerErrorf(
+		return controllererrors.NewErrorWithInnerErrorf(
 			controllererrors.ErrorExportingWithOpenSSL,
 			err,
 			"unable to export using openssl, error=[%v], stderr=[%v]", err, cerr.String(),
 		)
-		glog.Error(e.Error())
```
3quanfeng, line 146: keep the logging?

3quanfeng (approved):

3quanfeng (changes requested):
handle logging properly

3quanfeng (approved):

PR #1073, opened 2020-11-27
helm: Small adjustment in the values.yaml template to generate the artifact correctly

jullianoacqio:
… correctly

## Checklist
- [X] The title of the PR is clear and informative

## Description
This is just a small correction so when the values.yaml artifact is generated from the values-template.yaml template we don't have problems on the chart.
As you can see, this chart was generated with the appgw property incorrectly: https://appgwingress.blob.core.windows.net/ingress-azure-helm-package/ingress-azure-1.3.0-rc1.tgz


akshaysngupta (approved):
Thanks for contributing!

PR #1125, opened 2021-02-03
fix: generate default listener with private ip when specified in helm

akshaysngupta:
## Checklist

## Description

This PR fixes a bug in listener generation where a listener is generated with public IP even when helm config specifies to only use private IP. This is a problem where intent is to make AppGateway intranet accessible only.


pkg/environment/environment.go
```diff
@@ -179,7 +181,7 @@ func GetEnv() EnvVariables {
 		AuthLocation:                os.Getenv(AuthLocationVarName),
 		IngressClass:                os.Getenv(IngressClass),
 		WatchNamespace:              os.Getenv(WatchNamespaceVarName),
-		UsePrivateIP:                os.Getenv(UsePrivateIPVarName),
+		UsePrivateIP:                usePrivateIP,
```
3quanfeng, line 184: maybe enablePrivateIP? appgw doesn't seem to  support use privateIP only

akshaysngupta: usePrivateIP should mean the same

akshaysngupta: also, here we force all rules to be tried to `usePrivateIP`.
`enablePrivateIP` might confuse users since AGIC doesn't enable/add private to AppGateway. It barely uses it.

3quanfeng (approved):

PR #1271, opened 2021-09-23
fix: update AAD and Azure endpoints using Cloud from cpConfig

akshaysngupta:
## Checklist

## Description

* This issue is affecting AGIC addon in non-public cloud regions like China / US regions
* It affects addon when cluster is deployed in service principal mode.
* In this mode, AGIC uses the `clientId` and `clientSecret` fromm the cloud provider config.
* In this case, AGIC uses the standard `auth.NewClientCredentialsConfig` Auth library call which by default to public endpoints for AAD and ARM.
* This fails in non-public clouds.

In this fix, we use the `Cloud` name provided in the cloud provider config to initialize the correct endpoints.

## Tests
This was tested in the mooncake AKS deployment.

Error without the fix:
image

After the fix, AGIC is able to successfully get a token and update AppGateway
image

### DIFF
--- a/pkg/azure/auth.go
+++ b/pkg/azure/auth.go
@@ -10,6 +10,7 @@ import (
 
 	n "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-03-01/network"
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


pkg/azure/auth.go
```diff
@@ -40,6 +41,12 @@ func getAuthorizer(authLocation string, useManagedidentity bool, cpConfig *Cloud
 	if !useManagedidentity && cpConfig != nil {
 		klog.V(1).Info("Creating authorizer using Cluster Service Principal.")
 		credAuthorizer := auth.NewClientCredentialsConfig(cpConfig.ClientID, cpConfig.ClientSecret, cpConfig.TenantID)
+
+		// Set active directory endpoint using environment
+		azureEnv, _ := azure.EnvironmentFromName(cpConfig.Cloud)
+		credAuthorizer.AADEndpoint = azureEnv.ActiveDirectoryEndpoint
```
3quanfeng, line 47: does NewClientCredentialsConfig have any constructor which allows u to initialize endpoint directly?

akshaysngupta: no other constructor that takes clientId and secret as an input. There is one that takes from environment.

3quanfeng (approved):

PR #1272, opened 2021-09-23
fix: add type: object and add api-approved

akshaysngupta:
## Checklist

## Description

E2E test were failing in installing the Prohibited CRD changes as following fields were missing:

image

This was not caught in the previous PR as I didn't delete CRD before applying the CRD. Upgrading the CRD didn't catch these issues.

Adding `api-approved.kubernetes.io` and point it to this PR.
Since prohibited target CRD is using `*.k8s.io` as API group, API server is rejecting the CRD. This breaking change was released after prohibited target CRD was released as described in PR `https://github.com/kubernetes/enhancements/pull/1111`.

All other AGIC CRDS use `*.azure.io` which will not face this issue.


crds/AzureIngressProhibitedTarget-v1-CRD-v1.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
```
3quanfeng, line 2: it seems that our e2e infra doesn't clean up some of the resources, e.g. the prohibited crd object, we might need another PR  to fix the issue if any,

3quanfeng (approved):
Please log another work item to fix the resource cleanup for E2E

Issue #1257, opened 2021-09-06
Multiple AKS behind one AppGW


Exzelsios:
**Describe the bug**
We have several applications that each run on their own AKS. We want to run them all behind the same app gateway, but on different routes. For example, in the following logs we will have two applications: gbpoc running behind /name/ and testapp running behind /testapp/. The AGICs get deployed via Helm, being configured with the "shared" parameter set to true. 

So we have one application up and running with the AppGW (in this case gbpoc with the route /name/). The AGIC configures the listener, routing rule, backend pool etc. and creates the health probe as well. This works fine until we try to deploy the second app (testapp with route /testapp/) with its own AKS. 

Without the prohibit target CRD, this will overwrite our first app, making it unusable. 

With the "prohibit-all" CRD, obviously nothing happens, since the second AGIC isn't allowed to make changes to the AppGW. 

If we configure our own CRD i.e.:
```cat <<EOF | kubectl apply -f -
apiVersion: "appgw.ingress.k8s.io/v1"
kind: AzureIngressProhibitedTarget
metadata:
  name: test-custom-prohibitions
spec:
  paths:
    - /name/*
EOF
```
then our first app isn't overwritten, but the second app doesn't generate anything - kind of similar to the "prohibit-all" CRD case. We'd need to manually configure the listeners, backend probes etc. 

Is it possible to let multiple AGICs configure the AppGW without overwriting everything or manually creating/configuring the AppGW? 

In our logs, there seems to be a problem with routing rules. Is that related to it, did we misconfigure something?

If more information is needed, don't hesitate to ask.

**To Reproduce**
Two (or more) AKS
Deploy one app with an AGIC in one AKS. Configure the ProhobitTarget CRD to the path of the other app.
Once it's up and running with the AppGW, deploy a second app on a second AKS with AGIC. Ideally it would configure both applications automatically, but that's not what is happening. 

**Ingress Controller details**

Ingress Controller Pod Events: 
```Events:
  Type     Reason                     Age                From                       Message
  ----     ------                     ----               ----                       -------
  Normal   Scheduled                  19m                default-scheduler          Successfully assigned ingress-controller/agic-ingress-azure-5d9f6bbcdf-8cnw6 to aks-worker-15562235-vmss000000
  Normal   Pulling                    19m                kubelet                    Pulling image "mcr.microsoft.com/azure-application-gateway/kubernetes-ingress:1.4.0"
[...]

PR #1278, opened 2021-09-29
fix(prohibited target): incorrect merge when rules being merged reference the same path map 

akshaysngupta:
## Checklist

## Description

This PR fixes a bug in prohibited target.

When AGIC merges two routing rule together - existing and generated one - it also merged the attached URL path maps.
After merging AGIC deletes the second path map and only keeps the first one. This causes a bug when both first path map and second path map have the same `name`.

When path maps have the same name, they have already been merged when AGIC processes the path map before processing request routing rule. So, there is no need to merge the path maps again as it is already having the new rules.

## Fixes

Fixes https://github.com/Azure/application-gateway-kubernetes-ingress/issues/1257

### DIFF
--- a/pkg/brownfield/routing_rules_test.go
+++ b/pkg/brownfield/routing_rules_test.go
@@ -12,6 +12,7 @@ import (
 	ptv1 "github.com/Azure/application-gateway-kubernetes-ingress/pkg/apis/azureingressprohibitedtarget/v1"
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/tests"
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/tests/fixtures"
+	n "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-03-01/network"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/brownfield/routing_rules_test.go
```diff
@@ -101,4 +102,104 @@ var _ = Describe("Test blacklist request routing rules", func() {
...
+					ruleDefault,
+					rulePathBased1,
+				},
+				[]n.ApplicationGatewayRequestRoutingRule{
+					rulePathBased2,
+					rulePathBased3,
+				})
+
+			Expect(len(requestRoutingRules)).To(Equal(5))
+			Expect(len(*appGw.URLPathMaps)).To(Equal(4))
```
3quanfeng, line 148: Is it 4 before doing the merge?

akshaysngupta: yes and it should not decrease as we should retain all the url path maps

3quanfeng (dismissed):

Issue #462, opened 2019-08-06
Support rewriting HTTP headers


landro:
We'd like to be be able to define HTTP header rewrite rules (ref https://docs.microsoft.com/en-us/azure/application-gateway/rewrite-http-headers) using Kubernetes objects, either as Ingress annotations or as CRD. 

A typical use case is the need to set certain security heades like for instance Strict-Transport-Security, Server, X-Content-Type-Options, X-Frame-Options, X-XSS-Protection. The link above contains several additional use cases. 

Currently, these rewrite rules can not be set using the app gw ingress controller infrastructure, and risk being wiped if set using other mechanisms.

Related issue:
https://github.com/Azure/application-gateway-kubernetes-ingress/issues/437

akshaysngupta:
Thanks for suggesting! This is definitely on our road map but we are yet to assign an ETA.

landro:
Thanks for commenting on this, @akshaysngupta ! I realise you don't have an ETA yet, but what would be your best guess? Are we talking Q4 2019 or Q1 2020? I'm asking, because we're planning several major migrations to AKS in the coming months, and it would be kinda nice to know how much effort we should put in a workaround.

landro:
Probably makes sense to have a look at how the standard nginx ingress controller does this kind of thing (ConfigMap config): https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/

asridharan:
@landro instead of going the `ConfigMap` route we are probably going to define a CRD to have more strongly typed configurations. 

As @akshaysngupta mentioned while this feature is on our road-map we probably won't get till Q4'19.

rnarayana:
Hi, I also want to add the headers for security purposes. Is the workaround right now to add these at the backend services level?

landro:
That's what we do, @rnarayana . Generally speaking, you could either:

- add a "reverse proxy" sidecar (apache, ngingx, haproxy etc) and rewrite headers there
- rewrite headers in the application itself

evenh:
Any update on this functionality?

mscatyao:
Hey @evenh - this is still on our backlog, unfortunately we don't have an ETA for this yet as we're working a couple other features right now.

ahojman:
Hello everyone! 🙋🏻‍♂️- Upvoting this feature, as it's being also requested around! @mscatyao we'll much appreciate if this can be placed little bit up on the priority list.

[...]

PR #1287, opened 2021-10-11
rewrite: Add support for rewrite rule set assignment

### LINKED ISSUE #1003: Enable the annotations to modify App gateway "rewrite"  section from AGIC.
We have a need to use the section called "rewrite' inside an application gateway , but every single update done by AGIC in yaml , will overwrite the values listed in the application gateway.

Which annotations or methods can be used to update from AGIC the section rewrite of the application gateway? 

We have an alternative which is NGINX https://docs.microsoft.com/en-us/azure/application-gateway/tutorial-http-header-rewrite-powershell#specify-your-http-header-rewrite-rule-configuration  , but , can the same be done using AGIC?

beandrad:
So that rewrite rule sets predefined in the app gateway can be assigned
to routing rules by adding an annotation in the corresponding Ingress.

Fixes #1003.
Related to #462.

## Checklist

### DIFF
--- a/pkg/appgw/requestroutingrules.go
+++ b/pkg/appgw/requestroutingrules.go
@@ -68,7 +68,8 @@ func (c *appGwConfigBuilder) getRules(cbCtx *ConfigBuilderContext) ([]n.Applicat
 	httpListenersMap := c.groupListenersByListenerIdentifier(cbCtx)
 	pathMap := []n.ApplicationGatewayURLPathMap{}
 	var requestRoutingRules []n.ApplicationGatewayRequestRoutingRule
-	for listenerID, urlPathMap := range c.getPathMaps(cbCtx) {
+	urlPathMaps := c.getPathMaps(cbCtx)
+	for listenerID, urlPathMap := range urlPathMaps {
 		routingRuleName := generateRequestRoutingRuleName(listenerID)
 		httpListener, exists := httpListenersMap[listenerID]
 		if !exists {
@@ -97,6 +98,7 @@ func (c *appGwConfigBuilder) getRules(cbCtx *ConfigBuilderContext) ([]n.Applicat
 				rule.BackendAddressPool = nil
 				rule.BackendHTTPSettings = nil
 			}
+			rule.RewriteRuleSet = urlPathMap.DefaultRewriteRuleSet
 		} else {
 			// Path-based Rule
 			rule.RuleType = n.ApplicationGatewayRequestRoutingRuleTypePathBasedRouting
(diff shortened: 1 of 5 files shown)


pkg/appgw/requestroutingrules.go
```diff
@@ -146,16 +154,18 @@ func (c *appGwConfigBuilder) noRulesIngress(cbCtx *ConfigBuilderContext, ingress
...
 			},
 		}
+		(*listenerIngress)[listenerID] = ingress
 	}
 }
 
-func (c *appGwConfigBuilder) getPathMaps(cbCtx *ConfigBuilderContext) map[listenerIdentifier]*n.ApplicationGatewayURLPathMap {
+func (c *appGwConfigBuilder) getPathMaps(cbCtx *ConfigBuilderContext) (map[listenerIdentifier]*n.ApplicationGatewayURLPathMap, map[listenerIdentifier]*networking.Ingress) {
 	urlPathMaps := make(map[listenerIdentifier]*n.ApplicationGatewayURLPathMap)
+	listenerIngress := make(map[listenerIdentifier]*networking.Ingress)
```
akshaysngupta, line 163: there can be many ingresses that tie up to the same listener. In following code, you will always take the last ingress and pass it up. in case the `ingress` order changes, it is possible that a different rewrite rule set is passed up to the rule. This will lead to an unstable configuration which might keep filling between two rule sets

akshaysngupta: can you check `getDefaultFromRule` which returns the rule level properties. It is currently calculating the backend pool and backend http settings. You can modify it to also calculate the rewrite rule set that will be applied at the rule level.

akshaysngupta (changes requested):
listener ingress map might lead to unstable configuration.

pkg/appgw/requestroutingrules.go
```diff
@@ -146,16 +148,18 @@ func (c *appGwConfigBuilder) noRulesIngress(cbCtx *ConfigBuilderContext, ingress
...
 
 func (c *appGwConfigBuilder) getPathMaps(cbCtx *ConfigBuilderContext) map[listenerIdentifier]*n.ApplicationGatewayURLPathMap {
 	urlPathMaps := make(map[listenerIdentifier]*n.ApplicationGatewayURLPathMap)
+	listenerIngress := make(map[listenerIdentifier]*networking.Ingress)
 	for ingressIdx := range cbCtx.IngressList {
 		ingress := cbCtx.IngressList[ingressIdx]
 
 		if len(ingress.Spec.Rules) == 0 {
-			c.noRulesIngress(cbCtx, ingress, &urlPathMaps)
+			c.noRulesIngress(cbCtx, ingress, &urlPathMaps, &listenerIngress)
```
akshaysngupta, line 162: `listenerIngress` is not needed any more. Can you please remove ?

akshaysngupta (approved):

mosbahmajed:
@akshaysngupta is that released with 1.5.0 ? do we have any documentation ?

bisaloa:
The docs are [here](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/annotations.md#rewrite-rule-set).

shubhamsre:
https://azure.github.io/application-gateway-kubernetes-ingress/annotations/

This should be updated as well

m-matharu:
@bisaloa , It seems that the documentation was updated but the feature still doesn't work. I raised a case with Microsoft support and they have confirmed that the annotation is not useable with 1.5.0 rc1 but might be released in 1.5.0. As usual, no gurantees.

jsoumelidis:
As @m-matharu mentioned, the feature is still not working. Waiting for it...

PR #1344, opened 2022-02-12
k8s: Add retry logic to get server version

### LINKED ISSUE #1342: AGIC silently bypassed the networking failure during the pod initialization
**Describe the bug**
In v1.22 cluster, when the pod v1.5.0-rc starts to discover server version, if there is networking issue like failed to connect to the apiserver, AGIC silently bypass it instead of raising exception. This may result the cache sync failed because the pod keeps looking for 'v1beta1.Ingress'
eg. https://github.com/Azure/application-gateway-kubernetes-ingress/blob/1.5.0-rc1/pkg/k8scontext/supported_apiversion.go#L41   
**To Reproduce**
1. deploy AGIC 
2. block the connection to apiserver from the AGIC pod via iptables
3. restart the container of AGIC
4. checking the logs we could find message like below. It should not use 'v1beta1' any more in v1.22. 
I0211 09:39:48.815944       1 reflector.go:255] Listing and watching ***v1beta1.Ingress** from pkg/mod/k8s.io/client-go@v0.21.2/tools/cache/reflector.go:167

**Ingress Controller details**
* Output of `kubectl describe pod <ingress controller`> . The <ingress controller> pod name can be obtained by running `helm list`. 
* Output of `kubectl logs <ingress controller>. 
* Any Azure support tickets associated with this issue.

Shuanglu:
## Checklist

## Description

Add retry logic and raise exception when there is issue getting server version of the cluster

## Fixes

#1342

### DIFF
--- a/pkg/k8scontext/supported_apiversion.go
+++ b/pkg/k8scontext/supported_apiversion.go
@@ -67,6 +67,6 @@ func SupportsNetworkingPackage(client clientset.Interface) bool {
 		klog.Errorf("unexpected error parsing running Kubernetes version: %v", err)
 		return false
 	}
-
+	klog.V(1).Infof("server version is: %s", runningVersion.String())
 	return runningVersion.AtLeast(version119)
 }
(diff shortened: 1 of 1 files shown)


akshaysngupta (approved):
@Shuanglu thanks for fixing this! LGTM

PR #1433, opened 2022-07-29
Drop duplicate hostnames from ingress hostname annotation

sean-jeffrey:
This will prevent Application Gateway from rejecting a rule change
where a user has accidentally provided the same hostname twice in
an ingress hostname annotation.

## Checklist

## Description
This PR makes the AGIC ignore duplicate entries in an ingress hostnames annotation extension;
which would otherwise cause an error in the Application Gateway.

### DIFF
--- a/pkg/annotations/ingress_annotations.go
+++ b/pkg/annotations/ingress_annotations.go
@@ -276,10 +276,12 @@ func GetHostNameExtensions(ing *networking.Ingress) ([]string, error) {
 	if err == nil {
 		var hostnames []string
 		for _, hostname := range strings.Split(val, ",") {
-			if len(hostname) > 0 {
-				hostnames = append(hostnames, strings.TrimSpace(hostname))
+			trimmed := strings.TrimSpace(hostname)
+			if len(trimmed) > 0 {
+				hostnames = append(hostnames, trimmed)
 			}
 		}
+
 		return hostnames, nil
 	}
 
--- a/pkg/appgw/configbuilder_test.go
+++ b/pkg/appgw/configbuilder_test.go
@@ -962,4 +962,24 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
 
 		})
 	})
+
+	Context("hostname extension includes hostname from config rule", func() {
+		ingressAnnotations := make(map[string]string)
+		ingressAnnotations[annotations.HostNameExtensionKey] = "foo.baz, foo.bar, foo.bar"
+		ingress := &networking.Ingress{
+			Spec: networking.IngressSpec{
+				Rules: []networking.IngressRule{
+					{
+						Host: "foo.baz",
+					},
+				},
+			},
+			ObjectMeta: metav1.ObjectMeta{Annotations: ingressAnnotations},
+		}
+
+		It("should not include the duplicate host in the listener", func() {
+			listener := generateListenerID(ingress, &ingress.Spec.Rules[0], n.ApplicationGatewayProtocolHTTP, nil, false)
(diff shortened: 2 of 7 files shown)


akshaysngupta:
https://github.com/Azure/application-gateway-kubernetes-ingress/blob/45e25f5d4a9d38f4f582c6f87a5ac1bb30fa0e4a/pkg/appgw/configbuilder.go#L250

It also requires fixing in this section. Maybe instead of fixing this in the `GetHostNameExtension`, we fix this when we generated the final list in the above code section.

pkg/annotations/ingress_annotations.go
```diff
@@ -274,12 +274,21 @@ func BackendProtocol(ing *networking.Ingress) (ProtocolEnum, error) {
...
-				hostnames = append(hostnames, strings.TrimSpace(hostname))
+			trimmed := strings.TrimSpace(hostname)
+			if len(trimmed) > 0 {
+				uniqueHostnames[trimmed] = nil
 			}
 		}
+
+		// convert the map to a slice
+		var hostnames []string
+		for key := range uniqueHostnames {
```
akshaysngupta, line 288: Go `map` is unordered and will lose the ordering that customer gave.

akshaysngupta (changes requested):
left 2 comments

pkg/annotations/ingress_annotations.go
sean-jeffrey: Ah my bad for not realizing the order was significant, I've changed it to preserve the order

pkg/appgw/configbuilder_test.go
```diff
@@ -962,4 +962,23 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
 
 		})
 	})
+
+	Context("hostname extension includes hostname from config rule", func() {
+		ingressAnnotations := make(map[string]string)
+		ingressAnnotations[annotations.HostNameExtensionKey] = "foo.baz"
```
akshaysngupta, line 968: "foo.baz, foo.bar, foo.bar"

pkg/appgw/configbuilder_test.go
```diff
@@ -962,4 +962,23 @@ var _ = Describe("Tests `appgw.ConfigBuilder`", func() {
...
+		ingress := &networking.Ingress{
+			Spec: networking.IngressSpec{
+				Rules: []networking.IngressRule{
+					{
+						Host: "foo.baz",
+					},
+				},
+			},
+			ObjectMeta: metav1.ObjectMeta{Annotations: ingressAnnotations},
+		}
```
akshaysngupta, line 978: nit: newline

pkg/utils/utils_test.go
```diff
@@ -72,5 +72,19 @@ var _ = Describe("Utils", func() {
 				Expect(len(RandStringRunes(0))).To(Equal(0))
 			})
 		})
+
+		Context("Test RemoveDuplicateStrings", func() {
+			It("Should remove duplicate strings", func() {
+				input := []string{"1", "1", "2", "3", "4", "3", "5", "1"}
```
akshaysngupta, line 78: empty input list

pkg/appgw/configbuilder_test.go
sean-jeffrey: inserted an empty line

pkg/utils/utils_test.go
sean-jeffrey: used DescribeTable to test both empty slices and nil slices are returned unchanged

akshaysngupta (approved):
LGTM

PR #1464, opened 2022-10-26
[helm] Support Workload Identity

jkroepke:
## Checklist

## Description

To support [Workload Identity](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview), the helm chart should support define service account labels/annotations or BYO service account

### DIFF
--- a/helm/ingress-azure/templates/_helpers.tpl
+++ b/helm/ingress-azure/templates/_helpers.tpl
@@ -48,7 +48,9 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 If release name contains chart name it will be used as a full name.
 */}}
 {{- define "application-gateway-kubernetes-ingress.serviceaccountname" -}}
-{{- if .Values.fullnameOverride -}}
+{{- if .Values.serviceAccount.name -}}
+{{- .Values.serviceAccount.name -}}
+{{- else if .Values.fullnameOverride -}}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
--- a/helm/ingress-azure/templates/serviceaccount.yaml
+++ b/helm/ingress-azure/templates/serviceaccount.yaml
@@ -1,9 +1,19 @@
+{{ if .Values.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
+  name: {{ template "application-gateway-kubernetes-ingress.serviceaccountname" . }}
   labels:
     app: {{ template "application-gateway-kubernetes-ingress.name" . }}
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-  name: {{ template "application-gateway-kubernetes-ingress.serviceaccountname" . }}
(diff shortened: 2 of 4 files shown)


akshaysngupta:
@jkroepke the annotation defined by workload identity was fixed and only needed input from the user is the clientID of the user assigned identity. Why should service account take multiple annotations as an input ? Is there any benefit to that ?

Also, AGIC would require the sidecar proxy for migration as azure identity sdk for go doesn't support using a federated identity as a credential.

jkroepke:
> @jkroepke the annotation defined by workload identity was fixed 

The master branch does not accept annotations on service accounts?

> Why should service account take multiple annotations as an input ? Is there any benefit to that ?

For 3rd party integration, for example Kyverno. Also the charts created via `helm create` allows to define multiple annotations on service account. From a helm chart maintainer perspective, there is no blocker to offer this.

> Also, AGIC would require the sidecar proxy for migration as azure identity sdk for go doesn't support using a federated identity as a credential.

Yes, I'm aware of this.

AnthonyDewhirst:
Any further support on this? I am just moving all the rest of our services over to workload identity from pod identity but the use of AGIC means that I will still have at least 1 component using podIdentity which, I would love to remove.

payquicker-jlegan:
> Any further support on this? I am just moving all the rest of our services over to workload identity from pod identity but the use of AGIC means that I will still have at least 1 component using podIdentity which, I would love to remove.

@AnthonyDewhirst. we are in the same boat. I went through the RC for 1.6 and I am not seeing where the comment from @akshaysngupta is handled. I do not want to enable pod identity after rewriting all of our deployments to use workload identity on new, production clusters, just for the AGIC if the bits are out there and even if we have to fork it for the time being.

AnthonyDewhirst:
Does the following mean that we can use a solution without the sidecar now:
https://dev.azure.com/AzureContainerUpstream/Azure%20Workload%20Identity/_build/results?buildId=68454&view=results

jkroepke:
No, Identity Workload is not nativly implement in the Go SDK

AnthonyDewhirst:
https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity#key-concepts #2 doesn't this imply that it does? The diagram calls this out that it does.
And the following was what prompted my previous comment even though I didn't like directly to it:
https://github.com/Azure/azure-workload-identity/blob/main/go.mod

Forgive me if I am missing something in the detail here.

jkroepke:
@AnthonyDewhirst The content is available only on main branch, but not available in the current released version https://github.com/Azure/azure-sdk-for-go/tree/sdk/azidentity/v1.2.0/sdk/azidentity

jbpaux:
So I guess in January when 1.3.0 will be released we'll be able to use it in AGIC ? 😊

akshaysngupta:
I will start working on this next week. Will shoot for a release candidate mid next month.

jkroepke:
> So I guess in January when 1.3.0 will be released we'll be able to use it in AGIC ? 😊

I guess in 2024, since yearly relase scheudle

akshaysngupta:
Here is a "WIP" PR to support workload identity. Please leave any questions/comments on the PR.

https://github.com/Azure/application-gateway-kubernetes-ingress/pull/1498

jkroepke:
Thanks @akshaysngupta  for providing a more native way!

Issue #1524, opened 2023-03-24
Support Azure CNI Overlay


jkroepke:
**Is your feature request related to a problem? Please describe.**
AKS offers a new [Azure CNI Overlay](https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay) as alternative to kubenet.

There is documented limitation that AGIC would not work with Azure CNI Overlay.

> You can't use Application Gateway as an Ingress Controller (AGIC) for an overlay cluster.

**Describe the solution you'd like**

I would like to AGIC with Azure CNI Overlay.

EppO:
AGIC adds the backend's pod IP addresses in the backend pool for a specific ingress as the Application Gateway can't use the Kubernetes service (which uses non-routable IP range outside of the cluster). Azure CNI overlay prevents the Application Gateway to reach the pods directly. I don't think it's a use case that is possible. Using a Kubernetes Service with Nodeport or LoadBalancer type maybe? that's an extra-hop but it might help the connectivity between the application gateway and the pods.

jkroepke:
@EppO 

AGIC works fine with kubenet which based on non routable IP range outside of the cluster. Exactly the same behavior. This can't blocker.

EppO:
what IPs are using in the App Gw's backend pool in case of kubelet? I'm curious how this works with NAT-ed pod IPs

jkroepke:
The Pod IPs.

On AKS with kubenet,  AKS is managing a Route Table. Each Node gets a /24 network form the POD CIDR and AKS will configure UDRs the subnet route table. Based on that, the. AppGW has the ability to route to the Pod IPs directly.

Ref:
* https://learn.microsoft.com/en-us/azure/aks/configure-kubenet#overview-of-kubenet-networking-with-your-own-subnet 
* https://azure.github.io/application-gateway-kubernetes-ingress/how-tos/networking/#with-kubenet

nicolasdesentryfy:
Is there no plan to support Application Gateway with Overlay ?
It's really odd because overlay is what is being recommended to use as network plugin method. 

Here is a conflict:
We are using AKS Overlay with Azure Firewall. In the documentation, Azure Firewall recommends to use Application Gateway to keep the source IP in a X-Forwarded-For. However, we cannot, since Overlay doesn't support Application Gateway.
https://learn.microsoft.com/en-us/azure/architecture/example-scenario/gateway/firewall-application-gateway

Thus, we are blocked. 
[...]

PR #1650, opened 2024-09-30
Overlay support in AGIC

akshaysngupta:
## Checklist

## Description

This pull request introduces overlay support in the AGIC, specifically addressing overlay changes. It also refactors the Kubenet CNI.

In Overlay cluster, AGIC will need to create an Overlay Extension Config in which it will specify the Subnet CIDR where the gateway is deployed. This config then gets processed by the AKS cluster to allow AppGW data plane to be able to reach the Pods.


helm/ingress-azure/templates/clusterrole.yaml
```diff
@@ -65,4 +65,22 @@ rules:
   verbs:
     - create
     - patch
+- apiGroups:
+    - "acn.azure.com"
+  resources:
+    - "overlayextensionconfigs"
```
snehachhabria, line 71: does the crd for this need to be installed on the cluster prior to use of this feature or install of AGIC (with this feature) ?

pkg/k8s/scheme.go
```diff
@@ -0,0 +1,23 @@
...
+	overlayextensionconfig_v1alpha1 "github.com/Azure/azure-container-networking/crd/overlayextensionconfig/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// NewScheme builds and returns k8s schemes used by ALB Controller.
+func NewScheme() (*runtime.Scheme, error) {
+	s := runtime.NewScheme()
+	sb := runtime.SchemeBuilder{
+		// Azure CNI CRDs
+		overlayextensionconfig_v1alpha1.AddToScheme,
```
snehachhabria, line 14: are these two crds present on the cluster by default ?

snehachhabria (approved):

pkg/cni/overlay.go
```diff
@@ -0,0 +1,94 @@
...
+
+		klog.Infof("Creating overlay extension config with subnet CIDR %s", subnetCIDR)
+		err := r.client.Create(ctx, &config)
+		if err != nil {
+			return errors.Wrap(err, "failed to create overlay extension config")
+		}
+		return nil
+	}
+
+	config.Spec.ExtensionIPRange = subnetCIDR
```
robogatikov, line 87: ExtensionIPRange field is going to be immutable. Will this cause issues?
https://github.com/Azure/azure-container-networking/pull/3056

cmd/appgw-ingress/main.go
```diff
@@ -199,18 +215,11 @@ func main() {
...
-		subnetID := *(*appGw.GatewayIPConfigurations)[0].Subnet.ID
-		routeTableID := azure.RouteTableID(azure.SubscriptionID(cpConfig.SubscriptionID), azure.ResourceGroup(cpConfig.RouteTableResourceGroup), azure.ResourceName(cpConfig.RouteTableName))
-
-		err = azClient.ApplyRouteTable(subnetID, routeTableID)
-		if err != nil {
-			klog.V(3).Infof("Unable to associate Application Gateway subnet '%s' with route table '%s' due to error (this is relevant for AKS clusters using 'Kubenet' network plugin): [%+v]",
-				subnetID,
-				routeTableID,
-				err)
+	if err := cni.ReconcileCNI(context.Background(), azClient, ctrlClient, env.AGICPodNamespace, cpConfig, appGw); err != nil {
```
robogatikov, line 218: @nairashu was also suggesting to validate at AGIC level that VNET ID is the same for OverlayExtensionConfig ExtensionIPRange cidr and for aks cluster node ip addresses.

pkg/cni/overlay.go
akshaysngupta: Thanks for sharing that. What happens if user delete and creates a config with the same name but different IP? Is that an issue ?

cmd/appgw-ingress/main.go
akshaysngupta: Ok. Do you what is the best way to get the Cluster VNET ID on the controller ?

pkg/k8s/scheme.go
akshaysngupta: These two are suppose to be present only on overlay clusters.

pkg/cni/overlay.go
robogatikov: If user deletes OEC custom resource, clean up logic will run that will delete all the programmed Network Containers. After that OEC CR will be deleted. So, I don't think that this is an issue

cmd/appgw-ingress/main.go
akshaysngupta: Will take this up as a separate pr

snehachhabria (approved):

michael-rechani:
Running into the same issue here - https://github.com/Azure/application-gateway-kubernetes-ingress/issues/1524.  Happy to test AGIC with AKS and Azure CNI overlay once the update has been made.

PR #1789, opened 2026-04-16
fix: detect overlay CNI via NetworkContainer type when NNC label is absent

jamescummins:
## Problem

During Kubenet→Overlay migration, AKS RP restarts AGIC and polls for an OverlayExtensionConfig (OEC). AGIC never creates it because `isClusterOverlayCNI()` only checks for the metadata label `kubernetes.azure.com/podnetwork-type: overlay` on NodeNetworkConfigs (NNCs).

DNC-RC creates NNCs with `status.networkContainers[].type` set to `overlay` but **does not set the metadata label** at that point. This causes a deadlock:

- AKS RP declares NNCs ready (checks `status.NetworkContainers` length)
- AGIC checks the metadata label → not found → returns `false` silently
- No OEC created → AKS RP times out → cluster enters `Failed` state

This has been reproduced live in a test cluster and matches a production customer issue (7+ day `Failed` state).

## Fix

Add a fallback check in `isClusterOverlayCNI()` that examines `status.networkContainers[].type` when the metadata label is not found. Also adds logging on the `false` path for debuggability.

### Before
\\\go
for _, nnc := range nodeNetworkConfigs.Items {
    if val, ok := nnc.Labels[PodNetworkTypeLabel]; ok && val == "overlay" {
        return true, nil
    }
}
return false, nil  // silent
\\\

### After
\\\go
// Check metadata label first (fast path)
for _, nnc := range nodeNetworkConfigs.Items {
    if val, ok := nnc.Labels[PodNetworkTypeLabel]; ok && val == "overlay" {
        return true, nil
    }
}

// Fallback: check NetworkContainer type on status (set by DNC-RC during migration)
for _, nnc := range nodeNetworkConfigs.Items {
    for _, nc := range nnc.Status.NetworkContainers {
        if nc.Type == nodenetworkconfig_v1alpha.Overlay {
            return true, nil
        }
    }
}
\\\

## Testing

- All 18 existing unit tests pass
- 2 new unit tests added:
  - Detects overlay when NNC has no label but `NetworkContainer.Type = Overlay`
  - Does NOT falsely detect overlay when `NetworkContainer.Type = VNET`
- Reproduced the bug in a live AKS cluster (Kubenet + AGIC addon → overlay upgrade)
- Confirmed NNCs have `type: overlay` in status but zero metadata labels during migration

## Files Changed

- `pkg/cni/overlay.go` — fallback detection + logging
- `pkg/cni/overlay_test.go` — 2 new test cases

### DIFF
--- a/pkg/cni/overlay.go
+++ b/pkg/cni/overlay.go
@@ -84,12 +84,29 @@ func (r *Reconciler) isClusterOverlayCNI(ctx context.Context) (bool, error) {
 		return false, errors.Wrap(err, "failed to list node network configs")
 	}
 
-	// if any NNCs are overlay then this cluster is using CNI Overlay
+	if len(nodeNetworkConfigs.Items) == 0 {
+		klog.V(3).Infof("No NodeNetworkConfigs found, cluster is not using overlay CNI")
+		return false, nil
+	}
+
+	// Check NNC metadata label first (set after migration completes)
 	for _, nnc := range nodeNetworkConfigs.Items {
 		if val, ok := nnc.Labels[PodNetworkTypeLabel]; ok && val == "overlay" {
 			return true, nil
 		}
 	}
+
+	// Fallback: check NetworkContainer type on status (set by DNC-RC during migration)
+	for _, nnc := range nodeNetworkConfigs.Items {
+		for _, nc := range nnc.Status.NetworkContainers {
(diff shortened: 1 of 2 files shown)


jamescummins:
@microsoft-github-policy-service agree company="Microsoft"

PR #1798, opened 2026-07-14
fix: wait for CRD caches before reconciliation

### LINKED ISSUE #1814: AGIC starts reconciling before AzureIngressProhibitedTarget cache is synced – protected App Gateway config can be overwritten in shared/brownfield mode
### Describe the bug
In shared (brownfield) mode AGIC relies on `AzureIngressProhibitedTarget` CRs to know which Application Gateway listeners/rules it must not touch. On startup, `k8scontext` starts the CRD informers but explicitly excludes them from `WaitForCacheSync` (`pkg/k8scontext/context.go`). Reconciliation can therefore begin while the prohibited-target cache is still empty, so the first reconcile loop sees zero prohibited targets and treats the whole gateway as AGIC-owned.

Result: settings that are protected by `AzureIngressProhibitedTarget` (manually managed listeners, rules, backend pools) can be removed/overwritten during the first reconciliation after a pod (re)start. Once the cache catches up the protection works again, but the damage from the first pass is already applied to the gateway.

### To Reproduce
1. Deploy AGIC with `appgw.shared=true` and at least one `AzureIngressProhibitedTarget` covering a manually configured listener/rule on the Application Gateway.
2. Restart the AGIC pod (or scale 0→1) on a cluster/API server where the CRD list call is slow enough that reconciliation starts before the informer has synced.
3. Observe the first reconciliation: the prohibited-target set is empty and the protected configuration is included in the generated gateway config (visible in the logs / resulting ARM update).

### Expected behavior
AGIC should not reconcile until all enabled informer caches, including the CRD informers, have synced – the same way it already waits for Ingress, Service, Endpoints, etc.

### Proposed fix
PR #1798 (`fix: wait for CRD caches before reconciliation`) addresses this: CRD caches are only skipped from `WaitForCacheSync` when `omitCRDs` is set (unit tests), and are awaited in the production controller. Production code change is 2 lines; a regression test with a gated informer verifies that startup blocks until the `AzureIngressProhibitedTarget` informer reports sync. All 39 specs in `./pkg/k8scontext` pass.

### Environment
- AGIC version: 1.9.x (issue present on current master)
- Mode: shared / brownfield (`appgw.shared=true`)

@sean-jeffrey – would you be able to take a look at #1798? It is a small, test-covered change that only affects startup ordering.

okbeo:
## Checklist

## Description

AGIC currently starts CRD informers but always excludes them from `WaitForCacheSync`. During startup, the worker can therefore reconcile Application Gateway before the `AzureIngressProhibitedTarget` cache is populated.

With shared/brownfield mode enabled, AGIC can temporarily see no prohibited targets and overwrite protected Application Gateway configuration before the CRD create event is processed.

This change:

- honors the existing `omitCRDs` flag when deciding whether CRD informers participate in initial cache synchronization
- waits for `AzureIngressProhibitedTarget` and other enabled CRD caches before production reconciliation starts
- adds a regression test verifying that production startup remains blocked until the prohibited-target informer reports successful synchronization

The existing `omitCRDs` parameter remains `true` in unit tests and `false` in the production controller.

### Testing

`go test -tags unittest ./pkg/k8scontext -count=1`

All 39 specs pass.

## Fixes
Fixes #1814

- honors the existing `omitCRDs` flag when deciding whether CRD informers participate in initial cache synchronization
- waits for `AzureIngressProhibitedTarget` and other enabled CRD caches before production reconciliation starts

### DIFF
--- a/pkg/k8scontext/context.go
+++ b/pkg/k8scontext/context.go
@@ -229,8 +229,10 @@ func (c *Context) Run(stopChannel chan struct{}, omitCRDs bool, envVariables env
 	for _, informer := range sharedInformers {
 		go informer.Run(stopChannel)
 		// NOTE: Delyan could not figure out how to make informer.HasSynced == true for the CRDs in unit tests
-		// so until we do that - we omit WaitForCacheSync for CRDs in unit testing
-		if _, isCRD := crds[informer]; isCRD {
+		// so until we do that - we omit WaitForCacheSync for CRDs in unit testing only.
+		// Production must wait for CRD caches before reconciliation starts. Otherwise, an empty
+		// AzureIngressProhibitedTarget cache can make AGIC overwrite protected brownfield config.
+		if _, isCRD := crds[informer]; omitCRDs && isCRD {
 			continue
 		}
 		hasSynced = append(hasSynced, informer.HasSynced)
--- a/pkg/k8scontext/k8scontext_test.go
+++ b/pkg/k8scontext/k8scontext_test.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	testclient "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/annotations"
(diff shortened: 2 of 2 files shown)


okbeo:
@microsoft-github-policy-service agree [company="BEO IT AG"]

okbeo:
Reopen

okbeo:
@microsoft-github-policy-service agree company="BEO IT AG"

okbeo:
@microsoft-github-policy-service agree

PR #1806, opened 2026-08-21
Fix tombstone-handling panics in k8scontext delete handlers

### LINKED ISSUE #1147: AGIC not finding prohibited targets
**Describe the bug**
AGIC brownfield deployment, version 1.2.0, one app gw, 4 clusters. I have observed in one occasion that in one of the clusters, AGIC didn't find deployed prohibited targets thus erasing the app gw configuration, which obviously resulted in the other 3 clusters not being accessible anymore.

**To Reproduce**
Steps to reproduce the behavior: Apparently random occurrence.

**Additional details:**  Unfortunately the pod instance is gone already, so I don't have direct access to the logs anymore, but I can dig something up from Azure Log Analytics. As far as I can tell, it reads the app gw config alright, but then there is this: 
```
  1 mutate_app_gateway.go:85] Brownfield Deployment is enabled, but AGIC did not find any AzureProhibitedTarget CRDs  Disabling brownfield deployment feature.
```
The only real error I see in the logs is this, not sure if relevant:
```
1 client.go:227] Error getting route table '/subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Network/routeTables/<redacted>'
```
I can try to dig up more of the logs from Analytics, just let me know what to look for. The 4 cluster setup has been running fine for a few months already and some of the clusters get stuff deployed in them almost on a daily basis, and this is the first occurrence of this kind.  The prohibited targets have not been changed at all during this time.

karlskewes:
## Checklist

## Description

Fixes panics and silent drops when an informer delivers a delete tombstone instead of the typed object.

- unwrap tombstones in `deleteFunc`, `secretDelete`, `ingressDelete` before use
- forward the unwrapped object, not the tombstone, to the work queue
- drop events with no recoverable object instead of panicking or forwarding one
- fix `ingressDelete` to accept `extensions/v1beta1` Ingress tombstones too
- share one unwrap helper so all three handlers behave and log consistently
- add tests for the tombstone, nil-object, and wrong-type-object cases

`client-go`'s `DeltaFIFO` can hand `OnDelete` a tombstone instead of the typed object when it misses a delete event during a relist. Observed panic in production, matching several reports in #1147 going back to 2021:

```
panic: reflect: call of reflect.Value.Elem on struct Value [recovered]
	panic: reflect: call of reflect.Value.Elem on struct Value

goroutine 178 [running]:
	.../pkg/k8scontext/handlers.go:66 +0x71
	.../pkg/k8scontext/handlers.go:50 +0x2c
	created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 163
```

## Fixes

Fixes #1147

### DIFF
--- a/pkg/k8scontext/handlers.go
+++ b/pkg/k8scontext/handlers.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/Azure/application-gateway-kubernetes-ingress/pkg/events"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
 )
 
 type handlers struct {
@@ -48,6 +49,14 @@ func (h handlers) updateFunc(oldObj, newObj interface{}) {
 }
 
 func (h handlers) deleteFunc(obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		if tombstone.Obj == nil {
+			klog.Errorf("unable to get object from tombstone with key %s", tombstone.Key)
+			return
+		}
+		obj = tombstone.Obj
+	}
+
 	ns := getNamespace(obj)
 	if _, exists := namespacesToIgnore[ns]; exists {
 		return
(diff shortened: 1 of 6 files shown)


karlskewes:
@microsoft-github-policy-service agree company="Grafana Labs"

